PR #3, opened 2023-07-20
Fix post-optimization address fixup for late instructions

hanno-becker:
While SLOTHY's constraint modelling approach supports both early and late instructions, we haven't so far used late instructions in any example. In part, that's because the post-optimization address offset fixup did not support this case, and manually adjusting the offsets is rather fiddly. 

This PR adjusts the post-optimization address offset fixup to work with late instructions as well.


hanno-becker:
Subsumed by #5

PR #4, opened 2023-10-10
Fix broken example link in README.md

ggorlen:


hanno-becker:
@ggorlen Indeed, thanks 👍

PR #26, opened 2024-01-19
Fix examples

dop-amin:
- Allow for `/* ... */`-style single-line comments
- Remove `.req` and `.unreq` directives from source when optimizing "seams" between loops
- Add missing `@` for `slothy:no-unfold` tags
- Remove symbolic NTT examples from `example.py`

### DIFF
--- a/slothy/helper.py
+++ b/slothy/helper.py
@@ -33,19 +33,32 @@ class SourceLine:
     """Representation of a single line of source code"""
 
     def _extract_comments_from_text(self):
-        # Match // and /* ... */ style comments with optionally preceeding code
-        match = re.search(
-            r"(?P<code>.*?)\s*(//|/\*)\s*(?P<comment>.*?)\s*(?:\*/|$)",
-            self._raw)
+        # Regex pattern for single-line comments (// and /* */)
+        comment_pattern = r'//.*?(?=(//|/\*|$))|/\*.*?\*/'
 
-        # Check if there are any comments
-        if not match:
+        # Find comment start
+        comment_start = min([s for s in
+                             (self._raw.find('//'), self._raw.find('/*'))
+                             if s != -1], default=len(self._raw))
+        # No comment start means either empty comment or none at all
+        if comment_start == len(self._raw):
             return
 
-        comment = match.group("comment")
+        # Extract code
+        code = self._raw[:comment_start].strip()
 
-        self._raw = match.group("code")
-        self._comments.append(comment)
+        # Extract comments
+        comments = []
+        for comment in re.finditer(comment_pattern, self._raw):
+            comment_text = comment.group()
+            # Remove comment indicators
+            if comment_text.startswith('//'):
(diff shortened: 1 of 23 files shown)


slothy/helper.py
```diff
@@ -33,11 +33,19 @@ class SourceLine:
     """Representation of a single line of source code"""
 
     def _extract_comments_from_text(self):
-        if not "//" in self._raw:
+        # Match // and /* ... */ style comments with optionally preceeding code
```
hanno-becker, line 36: This is a slight change of behaviour for lines of the form `text // comment A //comment B`, isn't it? Previously, we would have extracted `comment A` and `comment B` separately. Can you retain that?

dop-amin: @hanno-becker Thanks for pointing this out. 
Commit b9e5a805e023f2e066973e8d010201bdbfca66c7 should preserve the original behaviour. 
Find some test cases I ran below: 
```
Text: 'abc // def // ghi', Code: 'abc', Comments: ['def', 'ghi']
Text: 'abc /* def */ /* ghi */', Code: 'abc', Comments: ['def', 'ghi']
Text: 'just code', Code: 'just code', Comments: []
Text: 'just code//', Code: 'just code', Comments: ['']
Text: '// just comment', Code: '', Comments: ['just comment']
Text: '/* just comment */', Code: '', Comments: ['just comment']
Text: 'abc // mixed /* comment */ // styles', Code: 'abc', Comments: ['mixed', 'comment', 'styles']
Text: 'abc /* mixed // comment styles */', Code: 'abc', Comments: ['mixed // comment styles']
Text: 'abc // /* more mixing */', Code: 'abc', Comments: ['', 'more mixing']
``` 

Note that empty comments will be stripped away in `_trim_comments()` anyways.

slothy/helper.py
```diff
@@ -214,7 +235,9 @@ def reduce_source(src):
         assert SourceLine.is_source(src)
         for l in src:
             l.reduce()
-        return [ l for l in src if l.has_text() and not AsmHelper.is_alignment_directive(l) ]
+        return [ l for l in src if l.has_text() and
+                 not AsmHelper.is_alignment_directive(l) and
```
hanno-becker, line 239: I think the reason why I originally omitted this was that allocation directives could be used outside of the optimized region as well -- in which case the optimized source file would no longer be functional.

hanno-becker: Any chance we can keep this behaviour, and move directives out of optimization regions in the example files?

slothy/helper.py
hanno-becker: Thanks @dop-amin! I think this would be easier to read if the code handled `/* ... */`-style comments and `// ...`-style comments separately. E.g. you could first handle all `/* ... */` comments and then use the previous code for `// ...`.

hanno-becker:
Most of this PR has been merged as part of #35. The handling of `/* .. */` comments remains.

PR #34, opened 2024-03-15
Add --dry-run option to example.py

hanno-becker:
Prevents any reordering or renaming so forces the identity permutation. Useful to check that the overall mechanics of SLOTHY are working, including parsing of all examples.


mkannwischer:
Thanks! LGTM

PR #36, opened 2024-03-15
NTTs: remove ldr/str macros that are no longer needed

mkannwischer:
This PR simplifies the aarch64 NTTs by removing the ldr/str macros that are no longer needed because SLOTHY now understands the Neon instructions natively. 

@dop-amin, do you happen to have access to a big machine that you could use to re-optimize these and test if they all still work? There is a chance that some `qform_` definitions are missing. 

Side-note: We considered replacing `qform_XXX` with the new syntax `Q<XXX>` but that would result in the source assembly not being functional without running it through SLOTHY. So for now we'll stick with the qform `.req`s until we have a better idea.


hanno-becker:
@mkannwischer This is still relevant, right?

mkannwischer:
Not a prio to remove. Closing now.

PR #37, opened 2024-03-15
Add inverse NTTs for Kyber & Dilithium

dop-amin:
This PR introduces inverse NTTs for Kyber and Dilithium. 
The type of transposition and reduction is supposed to match the code from PQClean [1,2].

TODO:
- [x] Add optimized Dilithium invNTTs
- [ ] ~~Simplify syntax as in #36~~ (do this separately, we want to re-run every example for this)
- [ ] ~~Optional: Add `intt_kyber_1234_567`~~ (from experience, these variants perform not as well as the ones that merge more layers in the second merge)

[1] https://github.com/PQClean/PQClean/tree/8e221ae797b229858a0b0d784577a8cb149d5789/crypto_sign/dilithium3/aarch64
[2] https://github.com/PQClean/PQClean/tree/8e221ae797b229858a0b0d784577a8cb149d5789/crypto_kem/kyber768/aarch64


examples/naive/aarch64/intt_dilithium_1234_5678_manual_ld4.s
```diff
@@ -0,0 +1,512 @@
...
+        ldp x19, x20, [sp, #16*0]
+        ldp x21, x22, [sp, #16*1]
+        ldp x23, x24, [sp, #16*2]
+        ldp x25, x26, [sp, #16*3]
+        ldp x27, x28, [sp, #16*4]
+        ldr x29, [sp, #16*5]
+        add sp, sp, #(16*6)
+.endm
+
+.macro save_vregs // slothy:no-unfold
```
hanno-becker, line 201: Note that those still use the old tag notation, and should be replaced by `// @slothy: ...`

slothy/targets/aarch64/aarch64_neon.py
```diff
@@ -2821,6 +2821,11 @@ def make(cls, src):
             ]
         return obj
 
+class cmge(AArch64Instruction): # pylint: disable=missing-docstring,invalid-name
```
hanno-becker, line 2824: It might be worth introducing a parent class for compare instructions, like
```
class ASimdCompare(AArch64Instruction):
   """Parent class for ASIMD compare instructions"""

class cmge(ASimdCompare):
   ...
```

and then wherever the SWOG talks about all compare instructions, you can just define latency/throughput/units in one go. See the Neoverse N1 model for examples -- it makes the uArch model more readable.

slothy/targets/aarch64/apple_m1_firestorm_experimental.py
```diff
@@ -120,7 +120,7 @@ def get_min_max_objective(slothy):
      vqdmulh_lane,
      vmull, vmlal,
      vsrshr, vushr, vusra, vshl,
-     vand, vbic): ExecutionUnit.V(),
+     vand, vbic, cmge): ExecutionUnit.V(),
```
hanno-becker, line 123: Here you can now refer to `cmge` as `ASimdCompare` instead (assuming all compare instructions have similar performance characteristics).

slothy/targets/aarch64/apple_m1_icestorm_experimental.py
```diff
@@ -98,7 +98,7 @@ def get_min_max_objective(slothy):
      vqdmulh_lane,
      vmull, vmlal,
      vsrshr, vushr, vusra, vshl,
-     vand, vbic): ExecutionUnit.V(),
+     vand, vbic, cmge): ExecutionUnit.V(),
```
hanno-becker, line 101: As above.

slothy/targets/aarch64/apple_m1_firestorm_experimental.py
dop-amin: Thanks, I modified the models. However, if we have something like `is_qform_form_of`, then passing the parent class won't work. Do we want to keep it this way or do we want the function returned by is_qform_form_of to also return `True` on the child classes?

hanno-becker: @dop-amin Good point. I think `is_qform_of` should be able to handle parent classes, yes. I bet the way I did this in `aarch64_neon.py` is not terribly pythonic, but you could use `all_subclass_leaves()` here.

hanno-becker:
@dop-amin Are you going to investigate the CI failure or do you need help?

hanno-becker:
@dop-amin Is there [going to be] a sibling PR to PQAX as well adding tests for the inverse NTT?

dop-amin:
> @dop-amin Are you going to investigate the CI failure or do you need help?

Hi Hanno, I think the CI just times out because it takes too long to go through all the examples. Especially the ones using heuristics seem to take long because it involves so many individual calls to the solver. Do you have a suggestion on how to go about this? We could disable the CI for examples using heuristics.

dop-amin:
> @dop-amin Is there [going to be] a sibling PR to PQAX as well adding tests for the inverse NTT?

Yes, I've been planning to submit it for a couple of days [but now I finally did so](https://github.com/slothy-optimizer/pqax/pull/5). Thanks for reminding me.

hanno-becker:
> Hi Hanno, I think the CI just times out because it takes too long to go through all the examples. Especially the ones using heuristics seem to take long because it involves so many individual calls to the solver. Do you have a suggestion on how to go about this? We could disable the CI for examples using heuristics.

I am surprised by this because the dry run sets `functional_only=True`, `allow_renaming=False` and `allow_reordering=False` if I remember correctly -- this should not take long. Can you double-check that your scripts in `example.py` do not overwrite this?

example.py
```diff
@@ -1123,6 +1204,25 @@ def core(self, slothy):
         slothy.config.split_heuristic = False
         slothy.optimize_loop("layer5678_start")
 
+        slothy.config = Config(self.arch, self.target)
```
hanno-becker, line 1207: @dop-amin I think that's the issue -- this overwrites anything that is setup by the parent class, including the dry run parameters.

It's not ideal that one can't do what you did, but at the moment you have to make a copy of the config at the start of `core()`.

dop-amin: Thanks for pointing me to this. Should be fixed.

example.py
```diff
@@ -1110,6 +1165,8 @@ def core(self, slothy):
 
         slothy.config = conf.copy()
 
+        if self.timeout is not None:
```
hanno-becker, line 1168: @dop-amin Is this duplication deliberate?

example.py
```diff
@@ -1124,6 +1181,51 @@ def core(self, slothy):
...
+            infile += f"_{var}"
+        name += f"_{target_label_dict[target]}"
+
+        super().__init__(infile, name, rename=True, arch=arch, target=target, timeout=timeout)
+
+    def core(self, slothy):
+        conf = slothy.config.copy()
+
+        slothy.config.reserved_regs = [
+            f"x{i}" for i in range(0, 6)] + ["x30", "sp"]
```
hanno-becker, line 1200: @dop-amin Could we take the opportunity to mark `x18` as reserved for all [i]NTT examples?

examples/naive/aarch64/intt_dilithium_1234_5678_manual_ld4.s
```diff
@@ -0,0 +1,512 @@
...
+        vmls        \dst,  \src, modulus
+.endm
+
+.macro gs_butterfly_v a, b, root, root_twisted
+        vsub    tmp,    \a, \b
+        vadd    \a,    \a, \b
+        mulmod  \b, tmp, \root, \root_twisted
+.endm
+
+.macro mul_ninv dst0, dst1, dst2, dst3, dst4, dst5, dst6, dst7, src0, src1, src2, src3, src4, src5, src6, src7
```
hanno-becker, line 129: This is an optional improvement, but it may be better to address it while we're at it:

Ideally `mulmod` uses a temporary (perhaps fixed throughout the clean code) for intermediates, so that it works regardless of whether `a` and `b` alias. This would allow you to pass just in=out to `mul_ninv`. Things like overwriting `data0,..,data3` with the `mul_ninv` result of `datat4..7`, and thus effectively using `data0..3` as temporaries, should ideally not be present in clean code.

examples/naive/aarch64/intt_dilithium_1234_5678_manual_ld4.s
```diff
@@ -0,0 +1,512 @@
...
+        str_vo data8, in,  (8*(512/8))
+        str_vo data9, in,  (9*(512/8))
+        str_vo data10, in, (10*(512/8))
+        str_vo data11, in, (11*(512/8))
+        str_vo data12, in, (12*(512/8))
+        str_vo data13, in, (13*(512/8))
+        str_vo data14, in, (14*(512/8))
+        str_vo data15, in, (15*(512/8))
+
+        mul_ninv data8, data9, data10, data11, data12, data13, data14, data15, data0, data1, data2, data3, data4, data5, data6, data7
```
hanno-becker, line 487: Oof, excuse my ignorance, but can you remind me why we need this (algorithmically)?

examples/naive/aarch64/intt_dilithium_1234_5678_manual_ld4.s
```diff
@@ -0,0 +1,512 @@
...
+.text
+
+        .global intt_dilithium_1234_5678_manual_ld4
+        .global _intt_dilithium_1234_5678_manual_ld4
+
+.p2align 4
+modulus_addr:   .quad 8380417
+ninv_addr:      .quad 16382
+ninv_tw_addr:   .quad 4197891
+intt_dilithium_1234_5678_manual_ld4:
```
hanno-becker, line 251: Again an optional improvement, but better to do while we're at it:

Could you go through the clean [inv]NTTs and add a comment at their entry symbol describing their arguments and in what way (reduction / ordering) they deviate from a standard, fully-reduced NTT?

examples/naive/aarch64/intt_dilithium_1234_5678_manual_ld4.s
hanno-becker: Ah, ok. It's the usual scaling by `1/N`, but we only need to do it for half of the output because for the other half, it has already been incorporated into the twiddles.

Could you add a comment to that effect?

examples/naive/aarch64/intt_dilithium_123_45678.s
```diff
@@ -0,0 +1,533 @@
...
+.endm
+
+.macro mulmod dst, src, const, const_twisted
+        mul        \dst\().4s,  \src\().4s, \const\().4s
+        vqrdmulh   \src,  \src, \const_twisted
+        vmlsq       \dst,  \src, consts, 0
+.endm
+
+.macro montg_reduce a
+        srshr tmp.4S,  \a\().4S, #23
```
hanno-becker, line 62: Is this really a Montgomery reduction? It looks more like Barrett, leveraging somehow that `q` is very close to `2^23`.

What exactly is happening here, and what are the bounds guarantees?

I remember we talked about this at some point, but I cannot remember.

examples/naive/aarch64/intt_dilithium_123_45678.s
```diff
@@ -0,0 +1,533 @@
...
+        canonical_reduce data5, modulus_half, neg_modulus_half, t2, t3
+        canonical_reduce data6, modulus_half, neg_modulus_half, t2, t3
+        canonical_reduce data7, modulus_half, neg_modulus_half, t2, t3
+
+        str_vo data4, in, (4*(1024/8))
+        str_vo data5, in, (5*(1024/8))
+        str_vo data6, in, (6*(1024/8))
+        str_vo data7, in, (7*(1024/8))        
+
+        mul_ninv data4, data5, data6, data7, data0, data1, data2, data3
```
hanno-becker, line 517: It would be interesting to double-check if the `mul_ninv` and the `canonical_reduce` can be replaced by a refined Barrett multiplication the sense of https://eprint.iacr.org/2022/439.pdf.

example.py
dop-amin: No, that was copy-pasta.

examples/naive/aarch64/intt_dilithium_1234_5678_manual_ld4.s
dop-amin: I agree, changed the way the macro works for the {inv}NTTs.

examples/naive/aarch64/intt_dilithium_1234_5678_manual_ld4.s
dop-amin: That's a good point in general. I followed the way the current PQClean implementations handle the reductions sucht that we can have a fair comparison. 

Regarding the ordering: All of our NTTs on aarch64 (should) output in the canonical ordering.

examples/naive/aarch64/intt_dilithium_123_45678.s
dop-amin: To me, this appears to be a shortened version of `reduce32` from the reference implementation. As mentioned above, I followed the current state in PQClean. 
Anyhow, the naming is off.

hanno-becker: Could you adjust the macro names?

hanno-becker (approved):
LGTM, @dop-amin -- thank you very much for this work.

dop-amin:
> LGTM, @dop-amin -- thank you very much for this work.

Great, thanks for your feedback in the process!

PR #51, opened 2024-04-09
Add models and performance statistics for models exercised in CI

hanno-becker:
This may be useful in better understanding performance regressions such as https://github.com/google/or-tools/issues/4166.

### DIFF
--- a/example.py
+++ b/example.py
@@ -88,7 +88,7 @@ class Example():
     def core(self, slothy):
         slothy.optimize()
 
-    def run(self, debug=False, log_model=False, dry_run=False, silent=False, timeout=0):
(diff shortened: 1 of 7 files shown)


example.py
```diff
@@ -141,7 +141,8 @@ def run(self, debug=False, log_model=False, dry_run=False, silent=False, timeout
             slothy.config.variable_size = True
 
         if log_model is True:
-            slothy.config.log_model = f"{self.name}_model.txt"
+            slothy.config.log_model_dir = log_model_dir
```
hanno-becker, line 144:
```suggestion
            slothy.config.log_model_dir = log_model_dir
```
Test

PR #52, opened 2024-04-11
Add support for vmull2, ngc and sbcs with xzr right-op-input and xzr output

aqjune-aws:
This is a simple patch that adds support for three instructions: vmull2, ngc and sbcs with xzr right-op-input and xzr output.


aqjune-aws:
Just in case that the contribution needs additional steps, I will close this PR for a while.

PR #53, opened 2024-04-12
Add support for vmull2, ngc xzr sbcs xzrs, sbcs xzrs and cmn

aqjune:
This adds support for a few arithmetic instructions in AArch64.

The updated uarchs are Neoverse N1 and Cortex A55.

### DIFF
--- a/slothy/targets/aarch64/aarch64_neon.py
+++ b/slothy/targets/aarch64/aarch64_neon.py
@@ -1651,6 +1651,12 @@ class neg(AArch64BasicArithmetic): # pylint: disable=missing-docstring,invalid-n
     inputs = ["Xa"]
     outputs = ["Xd"]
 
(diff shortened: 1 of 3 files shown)


hanno-becker (commented):
Thank you @aqjune!

Do you mind updating the A55 model as well?

hanno-becker:
> For other arithmetic ops - what would be a good step to check these?

I don't follow yet -- can you elaborate?

aqjune:
> I don't follow yet -- can you elaborate?
I was wondering whether the subclass relation between `AArch*Instruction` and the new instructions could silently add cost models to microarchitectures other than Neoverse N1. I checked that it did not introduce that.

Also this PR is updated to support Cortex A55 now too.

aqjune:
Oops, the committer and author are different. I will quickly amend it...

hanno-becker (approved):
LGTM, thanks June!

PR #55, opened 2024-05-22
Add Armv7M support for Cortex-M4 and Cortex-M7

dop-amin:
This draft PR is a WIP of adding Armv7M support to slothy. We add the support with Keccak as a first example optimization target. 
The new uArch models for Cortex-M4 and Cortex-M7 are still very incomplete, the model for Cortex-M4 is lacking the modeling of `ldr` instructions pipelining their address and data phases. 
We further add the ability to resolve `.if ... .else ... .endif` constructs in the assembly source code (also nested). 

Open ToDos:

- [ ] Refine uArch models for M4 and M7
- [ ] Clean up and extend Arch model
- [ ] Add notion of back-to-back instructions instead of only instruction pairs by data dependency
- [ ] Add Kyber and Dilithium NTTs as a fruther example
- [ ] Feature idea: Automatically detect unnecessary spills to the stack ("between str and ldr, there are free registers available")
- [ ] Optional: Automatic selection between 16-bit and 32-bit variant of an instruction on Cortex-M7 (the 32-bit variant is subject to some constraint adding additional latency as per the M85 SWOG, while on the other hand, using the 15-bit variant may forbid certain register renaming oppurtunities).

Further ideas are welcome!


slothy/targets/arm_v7m/arch_v7m.py
```diff
@@ -0,0 +1,1070 @@
...
+
+from sympy import simplify
+
+llvm_mca_arch = "arm"  # TODO
+
+
+class RegisterType(Enum):
+    GPR = 1
+    FPR = 2
+    STACK_FPR = 3
```
hanno-becker, line 16: Are those stack and hint registers still needed with the advent of `// @slothy: reads/writes=....` annotations?

dop-amin: I think so, there are probably some more artifacts in the Arch model that stem from basing it on the aarch64 one. 
I am going to fix this when I continue work on the Arch model.

slothy/helper.py
hanno-becker, line 1: @dop-amin Could you hoist this into a separate PR please?

PR #57, opened 2024-06-03
DRAFT: Update OR-Tools 9.7->9.10

hanno-becker:


hanno-becker:
CI succeeds for 9.10 with modified parameters. However, it still comes with a huge performance loss for some CI tasks: For example, `examples_ntt_kyber_dilithium_neon_core` took 40min rather than 20min.

PR #67, opened 2024-07-09
Split heuristic performance estimation

### LINKED ISSUE #66: Estimate performance of final code when using the split heuristic
When using the split heuristic, SLOTHY does not emit a performance estimate (and graph) for the final optimization, hardening the analysis of the result.

- [ ] Enhance the split heuristic so that it emits the same performance graph as without split heuristic.

mkannwischer:
Resolves #66 

Still WIP. Still have to put this behind a config option and re-run all the optimization using the split heuristic.


hanno-becker:
@mkannwischer Ready?

mkannwischer:
No, not yet.

mkannwischer:
I added a configuration option (`split_heuristic_estimate_performance`) defaulting to `True`. 
I've verified now that the configuration option works fine for both `False` and `True`. 

Maybe it is not required to re-run all the examples at this point.

PR #81, opened 2024-07-30
Extend support for symbolic assembly via spilling

hanno-becker:
SLOTHY has long supported assembly which uses symbolic register names instead of architectural ones. This commit improves the compatibility of this feature with both (a) complex workloads using numerous symbolic registers, and (b) architectures offering only a small register file, by adding experimental support for a simplified form of register spilling.

### DIFF
--- a/slothy/core/config.py
+++ b/slothy/core/config.py
@@ -452,11 +452,12 @@ class Config(NestedPrint, LockAttributes):
 
     @property
     def has_objective(self):
-        """Indicates whether a secondary objective (beyond minimization of stalls)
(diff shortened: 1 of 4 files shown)


hanno-becker:
The CI is slowing down too much even if spilling is disabled. @mkannwischer As is, this is still good enough for you to explore its use for Keccak, but we need to restore the non-spill speed before merging.

hanno-becker:
After running CI multiple times and comparing with prior PRs, the runtimes seem to be within the normal variation. The PR is thus ready for review. @mkannwischer @dop-amin

PR #91, opened 2024-10-11
More general structure for loop parsing

dop-amin:
This PR adds a more generic strucutre to deal with loops: Subclasses of Loop implement a specific type of loop, for exmaple having a certain sequence of instructions at the end. The extraction works the same and is thus implemented in `Loop`, while the methods to produce the code differ. 

I am open to suggestions for how to improve this approach; for my usecases wrt. armv7m it worked fine like this.

### DIFF
--- a/slothy/core/slothy.py
+++ b/slothy/core/slothy.py
@@ -367,12 +367,11 @@ class Slothy:
         """Run fusion callbacks on loop body"""
         logger = self.logger.getChild(f"ssa_loop_{loop_lbl}")
 
-        pre , body, post, _, other_data = \
+        pre , body, post, _, other_data, loop = \
             self.arch.Loop.extract(self.source, loop_lbl)
-        (loop_cnt, _, _) = other_data
+        loop_cnt = other_data['cnt']
         indentation = AsmHelper.find_indentation(body)
 
-        loop = self.arch.Loop(lbl_start=loop_lbl)
         body_ssa = SourceLine.read_multiline(loop.start(loop_cnt)) + \
             SourceLine.apply_indentation(self._fusion_core(pre, body, logger), indentation) + \
             SourceLine.read_multiline(loop.end(other_data))
@@ -398,9 +397,9 @@ class Slothy:
 
         logger = self.logger.getChild(loop_lbl)
 
-        early, body, late, _, other_data = \
+        early, body, late, _, other_data, loop = \
             self.arch.Loop.extract(self.source, loop_lbl)
-        (loop_cnt, _, _) = other_data
+        loop_cnt = other_data['cnt']
 
         # Check if the body has a dominant indentation
         indentation = AsmHelper.find_indentation(body)
@@ -464,7 +463,6 @@ class Slothy:
             for i in range(1, num_exceptional):
                 optimized_code += indented(self.arch.Branch.if_equal(loop_cnt, i, loop_lbl_iter(i)))
 
-        loop = self.arch.Loop(lbl_start=loop_lbl)
         optimized_code += indented(preamble_code)
 
         if self.config.sw_pipelining.unknown_iteration_count:
--- a/slothy/core/slothy.py
+++ b/slothy/core/slothy.py
@@ -392,7 +392,7 @@ class Slothy:
 
         assert SourceLine.is_source(self.source)
 
-    def optimize_loop(self, loop_lbl, postamble_label=None):
+    def optimize_loop(self, loop_lbl, postamble_label=None, cnt_iter_factor=1):
         """Optimize the loop starting at a given label"""
 
         logger = self.logger.getChild(loop_lbl)
@@ -475,7 +475,7 @@ class Slothy:
         optimized_code += SourceLine.read_multiline(loop.start(
             loop_cnt,
             indentation=self.config.indentation,
-            fixup=num_exceptional,
+            fixup=num_exceptional * cnt_iter_factor,
             unroll=self.config.sw_pipelining.unroll,
             jump_if_empty=jump_if_empty))
         optimized_code += indented(kernel_code)
--- a/slothy/core/slothy.py
+++ b/slothy/core/slothy.py
@@ -392,7 +392,7 @@ class Slothy:
 
         assert SourceLine.is_source(self.source)
 
-    def optimize_loop(self, loop_lbl, postamble_label=None, cnt_iter_factor=1):
+    def optimize_loop(self, loop_lbl, postamble_label=None):
         """Optimize the loop starting at a given label"""
 
         logger = self.logger.getChild(loop_lbl)
@@ -475,9 +475,11 @@ class Slothy:
         optimized_code += SourceLine.read_multiline(loop.start(
             loop_cnt,
             indentation=self.config.indentation,
-            fixup=num_exceptional * cnt_iter_factor,
+            fixup=num_exceptional,
             unroll=self.config.sw_pipelining.unroll,
-            jump_if_empty=jump_if_empty))
+            jump_if_empty=jump_if_empty,
+            preamble_code=preamble_code,
+            postamble_code=postamble_code))
         optimized_code += indented(kernel_code)
         optimized_code += SourceLine.read_multiline(loop.end(other_data,
             indentation=self.config.indentation))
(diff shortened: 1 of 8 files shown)


slothy/core/slothy.py
```diff
@@ -464,7 +463,6 @@ def loop_lbl_iter(i):
             for i in range(1, num_exceptional):
                 optimized_code += indented(self.arch.Branch.if_equal(loop_cnt, i, loop_lbl_iter(i)))
 
-        loop = self.arch.Loop(lbl_start=loop_lbl)
```
hanno-becker, line 467: Was this just. dead?

dop-amin: No, but `loop` is now initialized as part of the extraction earlier ([line 400](https://github.com/slothy-optimizer/slothy/blob/f5fba205e985e2acf82b8043b57530abf334fff6/slothy/core/slothy.py#L400)) because the type of loop needs to be determined for the `start` method to be appropriate in the followig.

slothy/targets/aarch64/aarch64_neon.py
```diff
@@ -250,28 +227,63 @@ def extract(source, lbl):
...
                 continue
             if state == 1:
-                p = loop_end_regexp[0].match(l_str)
+                p = loop_end_regexp[loop_end_ctr].match(l_str)
                 if p is not None:
-                    reg0 = p.group("reg0")
-                    reg1 = p.group("reg1")
-                    imm = p.group("imm")
-                    state = 2
+                    if additional_data is None:
```
hanno-becker, line 232: Can you say what the intended logic is here? Are you expecting this to be set by the first end-regexp?

slothy/targets/arm_v81m/arch_v81m.py
```diff
@@ -159,25 +158,72 @@ def extract(source, lbl):
...
-        return pre, body, post, lbl, ("lr", "lr", 0)
+            raise FatalParsingException(f"Couldn't identify loop {lbl}")
+        return pre, body, post, lbl, additional_data
+
+    @staticmethod 
+    def extract(source, lbl):
+        for loop_type in Loop.__subclasses__():
+            try:
+                l = loop_type(lbl)
+                return l._extract(source, lbl) + (l,)
```
hanno-becker, line 190: The `... + (l,)` may merit a comment. What's happening here? While you are at it, could you document `_extract()` the return tuple given by `_extract()` in a brief docstring?

slothy/targets/arm_v81m/arch_v81m.py
```diff
@@ -159,25 +158,72 @@ def extract(source, lbl):
...
+            except FatalParsingException:
+                logging.debug("Parsing loop type '%s'failed", loop_type)
+                pass
+                
+        raise FatalParsingException(f"Couldn't identify loop {lbl}")
+
+class LeLoop(Loop):
+
```
hanno-becker, line 198: Nit: Add a docstring giving an example for the type of loop recognized here?

slothy/targets/aarch64/aarch64_neon.py
```diff
@@ -250,28 +227,63 @@ def extract(source, lbl):
...
+                l = loop_type(lbl)
+                return l._extract(source, lbl) + (l,)
+            except FatalParsingException:
+                logging.debug("Parsing loop type '%s'failed", loop_type)
+                pass
+                
+        raise FatalParsingException(f"Couldn't identify loop {lbl}")
+
+class SubsLoop(Loop):
+    def __init__(self, lbl="lbl", lbl_start="1", lbl_end="2", loop_init="lr") -> None:
```
hanno-becker, line 260: Nit: Add a docstring giving an example for the type of loop recognized here?

slothy/targets/arm_v81m/arch_v81m.py
```diff
@@ -36,6 +36,7 @@
 import inspect
 import re
 import math
+from abc import ABC, abstractmethod
```
hanno-becker, line 39: Can we hoist out the `Loop` class so we don't have to have copies in every architecture model? Maybe it could go to `helper.py` for now?

slothy/core/slothy.py
hanno-becker: Got it, thank you

hanno-becker (changes requested):
Thank you @dop-amin, overall this looks good -- better flexibility for loop forms has been an embarrassing gap for a while.

While you're at it, could you improve the doc a bit, and hoist the abstract `Loop` class somewhere where it can be shared between architecture models?

Also, would you mind adding minimal examples to `examples.py` which demonstrate each loop form, so we run them in CI?

slothy/targets/aarch64/aarch64_neon.py
dop-amin: Yes, the first end-regexp sets `additional_data`. I made this choice because that's usually the `subs` or `cmp` in which we can see the counter register and/or decrement. 
I realize that maybe I want to get `additional_data` for each instruction in the "end"-part and then merge the dictionaries in the end. I'll change that.

slothy/core/slothy.py
```diff
@@ -393,14 +392,14 @@ def fusion_loop(self, loop_lbl):
 
         assert SourceLine.is_source(self.source)
 
-    def optimize_loop(self, loop_lbl, postamble_label=None):
+    def optimize_loop(self, loop_lbl, postamble_label=None, cnt_iter_factor=1):
```
hanno-becker, line 395: Maybe add a word of documentation for `cnt_iter_factor`? And, while at it, for `postamble_label`? :)

slothy/targets/aarch64/aarch64_neon.py
```diff
@@ -168,16 +171,28 @@ def unconditional(lbl):
...
 
+class SubsLoop(Loop):
+    """
+    Loop ending in a flag setting subtraction and a branch.
+    
+    Example:
+    ```
+           loop_lbl:
+               {code}
+               sub[s] <cnt>, <cnt>, #1
```
hanno-becker, line 183: Actually, setting the flag is pointless here, or?

dop-amin:
@hanno-becker I think the PR now is a good starting point for abstraction of loop handling. However, I think there may pop up new cases in the future which will require slight tweaks, e.g., passing more/different inputs to the loop subclasses. I already pass some data where I know it will be useful from our experiments with Armv7m, esp. for more complicated loop constructions that check against a pointer that is modified inside the kernel.

dop-amin:
> Thank you @dop-amin, overall this looks good -- better flexibility for loop forms has been an embarrassing gap for a while.
> 
> While you're at it, could you improve the doc a bit, and hoist the abstract `Loop` class somewhere where it can be shared between architecture models?
> 
> Also, would you mind adding minimal examples to `examples.py` which demonstrate each loop form, so we run them in CI?

I think this should all be done by now.

slothy/targets/arm_v81m/arch_v81m.py
```diff
@@ -109,14 +112,25 @@ def default_reserved():
...
-
+class LeLoop(Loop):
+    """
+    Loop ending in a le instruction.
+    
+    Example:
+    ```
+           loop_lbl:
+               {code}
+               le <cnt>, loop_lbl
```
hanno-becker, line 123: The implementation is assuming that `cnt` increases by `1` per iteration, right?

hanno-becker: This can be broken through SW pipelining, or not, if the loop count increment is chosen as an early instruction?

hanno-becker: No, my fault -- this is already part of `LE`, I had forgotten.

slothy/targets/aarch64/aarch64_neon.py
```diff
@@ -169,119 +172,53 @@ def unconditional(lbl):
...
 
-        yield f"{indent}sub {reg0}, {reg1}, {imm}"
-        yield f"{indent}cbnz {reg0}, {lbl_start}"
-
-    @staticmethod
-    def extract(source, lbl):
-        """Locate a loop with start label `lbl` in `source`.
-
-        We currently only support the following loop forms:
+        yield f"{indent}sub {other['cnt']}, {other['cnt']}, {other['imm']}"
```
hanno-becker, line 219: This should use the same format that the original loop had? With/without flag, and potentially using `cbnz`, `bnz`, `bne`?

hanno-becker (approved):

slothy/targets/aarch64/aarch64_neon.py
hanno-becker: This is pre-existing, so let's not block the PR because of it.

PR #99, opened 2024-11-05
Split dry run by target to speed-up CI runs

mkannwischer:


hanno-becker (approved):
Good idea, thank you @mkannwischer

PR #103, opened 2024-11-22
Armv7m modeling

dop-amin:
This PR extracts some of the features from #61 and puts them on top of #91. 

More precisely, it provides:
1. An arch model for Armv7M(-E).
2. An uArch model for the Arm Cortex-M7.
3. Some small examples.


slothy/targets/arm_v7m/arch_v7m.py
```diff
@@ -0,0 +1,1882 @@
...
+from slothy.helper import SourceLine, Loop
+from sympy import simplify
+
+llvm_mca_arch = "arm"  # TODO
+
+
+class RegisterType(Enum):
+    GPR = 1
+    FPR = 2
+    STACK_FPR = 3
```
hanno-becker, line 17: Those are relicts from the AArch64 model, right? We don't seem to need them.

slothy/targets/arm_v7m/arch_v7m.py
```diff
@@ -0,0 +1,1882 @@
...
+        yield f"cmp {cnt}, #{val}"
+        yield f"bge {lbl}"
+
+    @staticmethod
+    def unconditional(lbl):
+        """Emit unconditional branch"""
+        yield f"b {lbl}"
+
+
+class VmovCmpLoop(Loop):
```
hanno-becker, line 134: Can you add a short example here, as in the other loop classes?

slothy/targets/arm_v7m/arch_v7m.py
```diff
@@ -0,0 +1,1882 @@
...
+                loop_cnt_reg = loop_cnt
+            for l in body_code:
+                if l.text == "":
+                    continue
+                inst = Instruction.parser(l)
+                if loop_cnt_reg.lower() == inst[0].addr and inst[0].increment is not None:
+                    inc_per_iter = inc_per_iter + simplify(inst[0].increment)
+        logging.debug(f"Loop counter {loop_cnt} is incremented by {inc_per_iter} per iteration")
+        # Check whether instructions modifying the loop count moved to
+        # pre/postamble and adjust the fixup based on that.
```
hanno-becker, line 165: Can we have some minimal examples exercising this? You can force the situation under test through `early/late` annotations.

slothy/targets/arm_v7m/arch_v7m.py
```diff
@@ -0,0 +1,1882 @@
...
+        new_fixup = 0
+        if postamble_code is not None:
+            new_fixup = 0
+            for l in postamble_code:
+                if l.text == "":
+                    continue
+                inst = Instruction.parser(l)
+                if loop_cnt in inst[0].args_in_out and inst[0].increment is not None:
```
hanno-becker, line 173: (How) Do we know that `cnt` has not been renamed? Are we implicitly assuming that the counter is only touched once, via an instruction which has it as an in-out argument (and thus, does not rename it)?

slothy/targets/arm_v7m/arch_v7m.py
```diff
@@ -0,0 +1,1882 @@
...
+                    new_fixup = new_fixup + simplify(inst[0].increment)
+
+        if new_fixup != 0 or fixup != 0:
+            yield f"{indent}push {{{self.additional_data['end']}}}"
+            yield f"{indent}vmov {self.additional_data['end']}, {self.additional_data['endf']}"
+
+        if new_fixup != 0:
+            yield f"{indent}sub {self.additional_data['end']}, {self.additional_data['end']}, #{new_fixup}"
```
hanno-becker, line 181: Why don't we merge the instructions for `new_fixup` and `fixup`?

mkannwischer: this is actually a bug here that has been fixed in the other branch. please update.

slothy/targets/arm_v7m/arch_v7m.py
dop-amin: Should be fixed.

slothy/targets/arm_v7m/arch_v7m.py
dop-amin: We disabled this for now, so the cases won't be caught.

slothy/targets/arm_v7m/arch_v7m.py
dop-amin: We disabled this for now, so the cases won't be caught.

slothy/targets/arm_v7m/arch_v7m.py
```diff
@@ -154,25 +154,27 @@ def start(self, loop_cnt, indentation=0, fixup=0, unroll=1, jump_if_empty=None,
...
-            new_fixup = 0
-            for l in postamble_code:
-                if l.text == "":
-                    continue
-                inst = Instruction.parser(l)
-                if loop_cnt in inst[0].args_in_out and inst[0].increment is not None:
-                    new_fixup = new_fixup + simplify(inst[0].increment)
-
-        if new_fixup != 0 or fixup != 0:
+        # new_fixup = 0
```
hanno-becker, line 179: @dop-amin Why is this merely commented out? Is going to come back? If not, can we just remove it?

mkannwischer:
There still seems to be some clean-up tasks needed, but let's merge the current state and do the rest as a follow-up PR.
@dop-amin - can you open an issue with the tasks that are still open?

PR #104, opened 2024-11-22
Add boundary_reserved_regs option

dop-amin:
Extracted from #61, was introduced originally in 9310140e4b007ecfd1027418ed75516a90b51140.

### DIFF
--- a/slothy/core/core.py
+++ b/slothy/core/core.py
@@ -2763,6 +2763,8 @@ class SlothyBase(LockAttributes):
             for r in self.config.sw_pipelining.boundary_reserved_regs:
                 if r not in alloc.keys():
                     continue
+                if producer.src.is_virtual is True or consumer.is_virtual is True:
+                    continue
                 self._Add(alloc[r] == False).OnlyEnforceIf(producer.src.pre_var, consumer.pre_var.Not())
 
         if self.config.sw_pipelining.pre_before_post:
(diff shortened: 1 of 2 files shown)


slothy/core/core.py
```diff
@@ -2756,6 +2756,16 @@ def _add_constraints_loop_optimization(self):
...
+                    
+        # Forbid emerging dependencies across the loop boundary if the
+        # loop boundary (that SLOTHY pretends does not exist) uses temporary registers
+        for (consumer,producer, _, _, _, _, alloc) in self._iter_dependencies_with_lifetime():
+            for r in self.config.sw_pipelining.boundary_reserved_regs:
+                if r not in alloc.keys():
+                    continue
+                if producer.src.is_virtual is True or consumer.is_virtual is True:
+                    continue
+                self._Add(alloc[r] == False).OnlyEnforceIf(producer.src.pre_var, consumer.pre_var.Not())
```
hanno-becker, line 2768: This needs a bit more thought and tests to be robust in general:
* There will also be cross-boundary dependencies from non-post to post
* There can be dependencies across loop iterations (say an accumulator being changed in every iteration) that need to be handled correctly as well

hanno-becker (changes requested):
This is not yet ready to be merged, there are more corner cases to be considered. We should also have some tests for this.

dop-amin:
I'm closing this as we should rather avoid this situation by pulling most/all of the loop boundary into the loop body as we did in #139 for Armv7M.

dop-amin:
I'm closing this as we should rather avoid this situation by pulling most/all of the loop boundary into the loop body as we did in #139  for Armv7M.

PR #105, opened 2024-11-22
Use Fusion CB for splitting

dop-amin:
Extracted from #61 as introduced via d1a075f75b78c05671b012b5730d066a06935f7e.

### DIFF
--- a/slothy/core/slothy.py
+++ b/slothy/core/slothy.py
@@ -412,7 +412,17 @@ class Slothy:
         return body
 
     def fusion_region(self, start, end, **kwargs):
-        """Run fusion callbacks on straightline code"""
+        """ Run fusion callbacks on straightline code replacing certain
+        instruction (sequences) with an alternative. These replacements are
+        defined in the architectural model by setting an instruction class'
+        global_fusion_cb.
+        
+        Args:
+            start: The label marking the beginning of the part of the code to
+                  apply fusion to.
+            end: The label marking the end of the part of the code to apply
+                  fusion to.
+        """
         logger = self.logger.getChild(f"ssa_{start}_{end}")
         pre, body, post = AsmHelper.extract(self.source, start, end)
 
@@ -423,7 +433,13 @@ class Slothy:
         assert SourceLine.is_source(self.source)
 
     def fusion_loop(self, loop_lbl, **kwargs):
-        """Run fusion callbacks on loop body"""
+        """Run fusion callbacks on loop body replacing certain instruction
+        (sequences) with an alternative. These replacements are defined in the
+        architectural model by setting an instruction class' global_fusion_cb.
+        
+        Args:
+            loop_lbl: Label of loop to which the fusions are applied to. 
+        """
         logger = self.logger.getChild(f"ssa_loop_{loop_lbl}")
 
         pre , body, post, _, other_data, loop = \
--- a/slothy/core/slothy.py
+++ b/slothy/core/slothy.py
@@ -382,7 +382,7 @@ class Slothy:
         dfgc = DFGConfig(c)
         return list(DFG(body, logger.getChild("dfg_find_deps"), dfgc).inputs)
 
-    def _fusion_core(self, pre, body, post, logger):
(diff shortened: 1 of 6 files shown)


slothy/targets/aarch64/aarch64_neon.py
```diff
@@ -3175,7 +3176,50 @@ def core(inst,t,log=None):
 
     return core
 
-veor.global_fusion_cb  = eor3_fusion_cb()
+# TODO: Test only...
+# veor.global_fusion_cb  = eor3_fusion_cb()
+
+def eor3_splitting_cb():
```
mkannwischer, line 3182: Please add a comment here what this is doing. 
Probably also worth pointing out that we dont actually use this in any crypto examples so far.

slothy/core/slothy.py
```diff
@@ -373,25 +374,12 @@ def fusion_loop(self, loop_lbl):
...
-        self.config.constraints.allow_reordering = False
-        self.config.sw_pipelining.enabled = False
-        self.config.split_heuristic = False
-        self.config.inputs_are_outputs = True
-        self.config.sw_pipelining.unknown_iteration_count = False
-        self.optimize_loop(loop_lbl)
-        self.config = c
-
-        assert SourceLine.is_source(self.source)
-
```
mkannwischer, line 394: Why was this removed?

slothy/core/slothy.py
```diff
@@ -342,28 +342,29 @@ def _fusion_core(self, pre, body, post, logger):
...
 
-    def fusion_region(self, start, end):
+    def fusion_region(self, start, end, **kwargs):
         """Run fusion callbacks on straightline code"""
         logger = self.logger.getChild(f"ssa_{start}_{end}")
         pre, body, post = AsmHelper.extract(self.source, start, end)
 
         body_ssa = [ SourceLine(f"{start}:") ] +\
-             self._fusion_core(pre, body, logger) + \
+             self._fusion_core(pre, body, post, logger, **kwargs) + \
```
mkannwischer, line 362: I am a bit confused here. the argument `post` wasn't introduced in this PR. Did this just never work?

slothy/core/slothy.py
```diff
@@ -342,28 +342,29 @@ def _fusion_core(self, pre, body, post, logger):
...
             [ SourceLine(f"{end}:") ]
         self.source = pre + body_ssa + post
         assert SourceLine.is_source(self.source)
 
-    def fusion_loop(self, loop_lbl):
+    def fusion_loop(self, loop_lbl, **kwargs):
```
mkannwischer, line 367: The functions `fusion_loop` and `fusion_region` are external APIs. 
Could you add documentation strings like we have for `optimize`, please? Does not have to be long.

mkannwischer (changes requested):
Thanks @dop-amin!
Some some questions and suggestions.

slothy/core/slothy.py
dop-amin: I think because it was never actually doing anything. In the `fusion_range`, it was never there in the first place. It makes no sense to call the actual optimizer in the fusion code, this is just supposed to perform replacements. 
But I think @hanno-becker made this change originally and may know better than I do.

slothy/core/slothy.py
dop-amin: I think so, this was not used at all if I recall correctly. But again, @hanno-becker made this change originally and may know more.

slothy/targets/aarch64/aarch64_neon.py
```diff
@@ -3175,7 +3176,50 @@ def core(inst,t,log=None):
 
     return core
 
-veor.global_fusion_cb  = eor3_fusion_cb()
+# TODO: Test only...
+# veor.global_fusion_cb  = eor3_fusion_cb()
```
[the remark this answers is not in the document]
dop-amin: I removed it and mention where we actually set the veor global_fusion_cb that these are two different options.

mkannwischer (approved):
Thanks for making the changes @dop-amin!

PR #108, opened 2024-11-25
Add equivalence tester

hanno-becker:
By default, SLOTHY includes a 'selfcheck' verifying that the DFGs
of input and output are isomorphic with respect to the rescheduling
found by SLOTHY. This selfcheck is very powerful and should catch most
modelling errors in SLOTHY.

However, some aspects of SLOTHY remain outside of the scope of the
DFG-based selfcheck:

- Address offset fixup

  This is a historically rather error prone feature to which the self-check
  is blind (since it relies on making the DFG blind to address increments).

- Loop boundaries

  If SW pipelining is enabled, the selfcheck checks equivalence of a
  _bounded unrolling_ of the input and output loops.

  Therefore, it would not catch issues with (a) the loop boundary, or
  (b) exceptional iterations for low loop counts.

While it is not SLOTHY's goal to provide formal guarantees of correctness
of the optimizations -- this should be gauged/confirmed through extensive test
and ideally formal verification -- development would be faster if bugs
in the aforementioned areas were caught before the developer integrates
and tests the SLOTHY output manually.

As a step towards catching more potential pitfalls early in the SLOTHY
workflow, this commit starts the development of an _equivalence tester_.

The idea of the equivalence tester is simple: Just run original and optimized
assembly with various random inputs, and check that they indeed lead to
equivalent outputs.

This commit implements this idea for AArch64 and Armv7-M, leveraging
LLVM-MC as an assembler and the unicorn-engine for emulation.

An initial limitation is support for loops: For now, the equivalence
checker tests bounded unrollings of loops, just like the selfcheck. That is
against the point of the equivalenc check, but it will be improved at a later
point, after some refactoring of the loop generation code.

The checker added in this commit is able to detect issues with address offset
fixup, however, which would previously have gone unnoticed by the selfcheck.

The new equivalence test is configured via config.selftest, which is enabled
by default, but silently skipped for unsupported platforms or target architectures.

### DIFF
--- a/slothy/targets/aarch64/aarch64_neon.py
+++ b/slothy/targets/aarch64/aarch64_neon.py
@@ -81,6 +81,22 @@ class RegisterType(Enum):
     def spillable(reg_type):
         return reg_type in [RegisterType.GPR, RegisterType.NEON]
 
+    @staticmethod
+    def callee_saved_registers():
+        return [f"x{i}" for i in range(18,31)] + [f"v{i}" for i in range(8,16)]
+
+    @staticmethod
+    def unicorn_link_register():
+        return UC_ARM64_REG_X30
+
+    @staticmethod
+    def unicorn_stack_pointer():
+        return UC_ARM64_REG_SP
+
+    @staticmethod
+    def unicorn_program_counter():
+        return UC_ARM64_REG_PC
+
     @cache
     @staticmethod
     def unicorn_reg_by_name(reg):
--- a/slothy/targets/aarch64/aarch64_neon.py
+++ b/slothy/targets/aarch64/aarch64_neon.py
@@ -40,17 +40,28 @@ similar to those used in the Arm ARM.
 import logging
 import inspect
 import re
+import os
 import math
+import platform
+import subprocess
 from enum import Enum
 from functools import cache
-
 from sympy import simplify
 
+from unicorn import *
+from unicorn.arm64_const import *
+
 from slothy.targets.common import *
-from slothy.helper import Loop
+from slothy.helper import Loop, LLVM_Mc
 
 arch_name = "Arm_AArch64"
+
 llvm_mca_arch = "aarch64"
+llvm_mc_arch = "aarch64"
+llvm_mc_attr = None
+
+unicorn_arch = UC_ARCH_ARM64
+unicorn_mode = UC_MODE_ARM
 
 class RegisterType(Enum):
     GPR = 1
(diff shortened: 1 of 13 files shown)


tests/aarch64/checker/README.md
```diff
@@ -0,0 +1,33 @@
...
+memory contents differ, the checker fails.
+
+### Configuration
+
+You can configure the checker by adjusting `harness_config.i`:
+
+* With `#define KERNEL_STACK_SIZE XXX` you can indicate that the kernels assume memory of size `XXX` starting from
+  `sp`. Note that you do not need to specify anything for the kernel(s) to use additional stack below `sp`. However,
+  it is assumed that the kernels restore the original value of `sp` (otherwise, the checker will likely segfault).
+* With `#define FPR_FIX_VALUE_X{0-30} VALUE` you can omit random assignment to a GPR and instead force a given
```
dop-amin, line 29: Probably a typo: Should be `{G,F}PR_FIX_VALUE_X{0-30}` or `GPR_FIX_VALUE_X{0-30}`, right?

tests/aarch64/checker/checker.c
```diff
@@ -0,0 +1,305 @@
+
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
```
dop-amin, line 4:
```suggestion
#include <string.h>
#include <stdint.h>
```

slothy/targets/aarch64/aarch64_neon.py
```diff
@@ -49,6 +51,111 @@ class which generates instruction parsers and writers from instruction templates
...
+         - result: Instance of Result class representing the SLOTHY run
+                   to be checked. This includes both the optimized and the
+                   original source code.
+         - address_gprs: GPR-indexed dictionary indicating which registers
+                   contain pointers to memory. The value indicates the size
+                   of the memory. For example, { "x0" : 1024 } would lead to
+                   X0 being set to the base of a 1024 byte buffer."""
+
+        # For now, we only support native AArch64 systems
+        if platform.machine() != "arm64":
```
dop-amin, line 89:
```suggestion
        if platform.machine() not in ["arm64", "aarch64"]:
```

On my Raspberry Pi 4, `platform.machine()` returns `aarch64`.

dop-amin (commented):
I really like how well the feature is integrated with slothy already. 

As you write, there's room for improvment regarding the abstraction of the approach. I think things like the number/types of registers etc. could be inferred from the Arch models automatically for example. 

I tested the checker with some small examples and also introduced mistakes on purpose that were caught. 
Apart from the suggested changes to the code, there have been no issues for me. 

Possibly, adding an example in `example.py` could be nice to make it complete.

slothy/targets/aarch64/aarch64_neon.py
hanno-becker: No longer relevant after the switch to unicorn.

tests/aarch64/checker/checker.c
hanno-becker: No longer relevant after the switch to unicorn.

tests/aarch64/checker/README.md
hanno-becker: No longer relevant after the switch to unicorn.

PR #110, opened 2024-11-28
Move export of CP-SAT model

mkannwischer:
Right now we are exporting the CP-SAT model after _solve() has returned.
However, at the end of _solve(), we add the found solution as hints to the model such that sub-sequent solving of the same model (with a different objective) run faster.
This resulted in the exported model not being the one that was being solved originally.

This commit moves the export.

### DIFF
--- a/slothy/core/core.py
+++ b/slothy/core/core.py
@@ -1475,8 +1475,6 @@ class SlothyBase(LockAttributes):
         self.result.success = self._solve()
         self.result.valid = True
 
-        # - Export (optional)
-        self._export_model()
 
         if not retry and self.success:
             self.logger.info("Booleans in result: %d", self._model.cp_solver.NumBooleans())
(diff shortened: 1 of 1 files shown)


slothy/core/core.py
```diff
@@ -3429,6 +3427,9 @@ def is_good_enough( cur, bound ):
 
         ok = self._model.cp_model.status in [cp_model.FEASIBLE, cp_model.OPTIMAL]
 
+        # - Export (optional)
```
hanno-becker, line 3430: It would be cleaner if this was before the call to `solve`?

hanno-becker (approved):
One nit, but not blocking. You decide @mkannwischer

slothy/core/core.py
mkannwischer: that doesn't work. it errors if you try that

hanno-becker: Why?

mkannwischer: more details: we printing stats about the solving in the export_model function. We would have to split that up if we want to export the model before solving.

mkannwischer: not doing that now

PR #114, opened 2024-12-05
Extend Armv7E-M architecture model

mkannwischer:
Small tweaks to the Armv7E-M model: 
- Add `smlatb` and `ldrb` with postincrement (and corresponding uArch model for M7)
- make barrel shift parsing more flexibile (can now write either `asr#1` or `asr #1` (before it was limited to the latter)


dop-amin (approved):
Looks good, thanks! The limitation about the barrel shift notation was indeed confusing.

dop-amin:
@mkannwischer If you wanted to, you could add an example that uses the added instructions just to have them "covered" by CI in some place.

mkannwischer:
We have the examples in the armv7m branch - I hope that we can merge all those soon.

PR #116, opened 2024-12-09
Small tweaks to the Armv7-M architecture model

mkannwischer:
This applies three small tweaks to the Armv7-M arch model 
- Ignore all optional spaces in instruction parsing
- Use the correct number of FPU registers (32 rather than 31)
- Make the ldm/stm register list/range parsing much more flexible

Please review each commit individually.


slothy/targets/arm_v7m/arch_v7m.py
```diff
@@ -1417,53 +1443,53 @@ def make(cls, src):
...
+        reg_list_type, reg_list = Armv7mInstruction._expand_reg_list(obj.reg_list)
+
+        obj.args_out = reg_list
         obj.num_out = len(obj.args_out)
         obj.arg_types_out = [RegisterType.GPR] * obj.num_out
-        obj.args_out_restrictions    = [[ f"r{i+j}"  for j in range(0, num_regs-obj.num_out)] for i in range(0, obj.num_out) ]
-        obj.args_out_combinations = [ ( list(range(0, obj.num_out)), [ [ f"r{i+j}" for i in range(0, obj.num_out)] for j in range(0, num_regs-obj.num_out) ] )]
+        available_regs = RegisterType.list_registers(RegisterType.GPR)
+        obj.args_out_combinations =  [ (list(range(0, obj.num_out)), [list(a) for a in itertools.combinations(available_regs, obj.num_out)])]
+        obj.args_out_restrictions = [ None for _ in range(obj.num_out)    ]
```
dop-amin, line 1466: Not precisely on your PR but: I think it would generally be helpful to have some documentation what `args_out_combinations` and `args_out_restrictions` actually are and how they're supposed to be set. On first sight, they can be quite confusing.

examples/opt/armv7m/armv7m_simple0_opt_m7.s
```diff
@@ -1,24 +1,42 @@
...
+                                      // Cycle bound:     14.0
+                                      // IPC bound:       1.71
+                                      //
+                                      // Wall time:     0.39s
+                                      // User time:     0.39s
+                                      //
+                                      // ----- cycle (expected) ------>
+                                      // 0                        25
+                                      // |------------------------|----
+        ldr r11,[r0,#4]               // *.............................
```
dop-amin, line 16: OK, here the spaces are gone for the `ldr`.

dop-amin (approved):
Looks fine to me, thanks for removing some of those historic limitations.

slothy/targets/arm_v7m/arch_v7m.py
```diff
@@ -940,242 +966,242 @@ def make(cls, src):
 
 # movs
 class movw_imm(Armv7mBasicArithmetic): # pylint: disable=missing-docstring,invalid-name
-    pattern = "movw <Rd>, <imm>"
```
hanno-becker, line 943: Nit: I think one shouldn't be forced to remove those spaces -- it should be the parent class interpreting a space at this point as an optional space of arbitrary size?

PR #123, opened 2024-12-15
Switch CI to MacOS runners

mkannwischer:
More of an experiment to see how much faster the MacOS runners are. 
Unfortuntely switching does not appear to be an option because Github's limits on concurrent MacOS jobs are much lower than for the ubuntu runners (5 vs 20 concurrent jobs).


mkannwischer:
Okay, seems like switchting this by itself does not improve performance. Nevermind.

PR #125, opened 2024-12-16
Refine selftest further

mkannwischer:
We encountered further issues with the selftest in the `armv7m` branch. This PR refines it further

1) When unicorn fails to run code, currently this fails silently and the selftest continues to run failing in the assertions checking the output resulting in misleading error messages. This is changed now to fail immediately with an appropriate eror message. 

2) The configuration option `selftest_address_gprs` was misnamed as addresses can also be contained in non-gpr-registers. I renamed it to `selftest_address_registers`

3) Previously the selftest would determine which registers are supposed to contain addresses  (and allocate memory
accordingly) by going through all memory operations and assuming that address registers used in those are the
inputs holding the addresses initially.

This assumption does not hold as
(1) addresses may be moved around
(2) addresses may actually be a result of a previous computation.

One example where this did not work was
```
vmov rX, sY
ldr rZ, [rX]
```

Here the selftest would assume that rX needs to be an address. However, sY is the register holding the address on entry of the code.

This commit refines the code determining the address registers by recursively going through the DFG and determining the _input_ registers that affect values being used as addresses.
Note that in the above examples rX would not be marked as an address, at it is not an input.

Hopefully this approach is more robust.

4) Allow accessing memory at negative offsets. Sometimes code has pointers that do not point at the start of the buffer, but somewhere in the middle. If those are then used with negative offsets, unicorn fails as the memory is potentially undefined. This is fixed by allocating a double-sized buffer and placing the pointer in the middle. 

5) The stack pointer is currently initialized to the top of the stack. However, if code is using the stack pointer (and the sp is not decremented as a part of the code snipped), then we get memory access errors from unicorn. This is fixed by allocating some stack space (the default buffer size) before calling the function. 

6) Building on top of #122, we also need to skip the selftest if symbolics are present in the preamble or postamble in the case of sw piplining. This PR adds the check in the same way as #122 did for the main body of the code.

### DIFF
--- a/slothy/core/core.py
+++ b/slothy/core/core.py
@@ -870,6 +870,21 @@ class Result(LockAttributes):
         iterations = 7
         if self.config.sw_pipelining.enabled is True:
             old_source, new_source = self.get_fully_unrolled_loop(iterations)
+
+            dfgc_preamble = DFGConfig(self.config, outputs=self.kernel_input_output)
+            dfgc_preamble.inputs_are_outputs = False
+            preamble_dfg = DFG(self.preamble, log, dfgc_preamble)
+
+            if preamble_dfg.has_symbolic_registers():
+                log.info("Skipping selftest as preamble contains symbolic registers.")
+                return
+
+            dfgc_postamble = DFGConfig(self.config, outputs=self.orig_outputs)
+            postamble_dfg = DFG(self.postamble, log.getChild("new_postamble"), dfgc_postamble)
+
+            if postamble_dfg.has_symbolic_registers():
+                log.info("Skipping selftest as postamble contains symbolic registers.")
+                return
         else:
             old_source, new_source = self.orig_code, self.code
 
--- a/slothy/core/core.py
+++ b/slothy/core/core.py
@@ -846,8 +846,8 @@ class Result(LockAttributes):
 
         log.info(f"Running selftest ({self._config.selftest_iterations} iterations)...")
 
-        address_gprs = self._config.selftest_address_gprs
-        if address_gprs is None:
+        address_registers = self._config.selftest_address_registers
+        if address_registers is None:
             # Try to infer which registes need to be pointers
             # Look for load/store instructions and remember addresses
             addresses = set()
@@ -862,8 +862,8 @@ class Result(LockAttributes):
             # just allocate a buffer of a configurable default size.
             log.info(f"Inferred that the following registers seem to act as pointers: {addresses}")
             log.info(f"Using default buffer size of {self._config.selftest_default_memory_size} bytes. "
-                     "If you want different buffer sizes, set selftest_address_gprs manually.")
-            address_gprs = { a: self._config.selftest_default_memory_size for a in addresses }
+                     "If you want different buffer sizes, set selftest_address_registers manually.")
+            address_registers = { a: self._config.selftest_default_memory_size for a in addresses }
 
         # This produces _unrolled_ code, the same that is checked in the selfcheck.
         # The selftest should instead use the rolled form of the loop.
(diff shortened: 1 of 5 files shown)


slothy/helper.py
```diff
@@ -1338,6 +1338,7 @@ def run_code(code, txt=None):
                 log.error("Failed to emulate code using unicorn engine")
                 log.error("Code")
                 log.error(SourceLine.write_multiline(code))
+                raise SelfTestException("Selftest failed: Unicorn failed to emulate code")
```
hanno-becker, line 1341: We should catch the exact exception and print the error information contained therein. Can be follow up.

slothy/core/dataflow.py
```diff
@@ -775,6 +775,30 @@ def has_symbolic_registers(self):
...
+        # if this is a virtual input instruction this is an actual input
+        # otherwise this is computed from other inputs
+        if isinstance(producer.inst, VirtualInputInstruction):
+            return set(producer.inst.args_out)
+        else:
+            # go through all predecessors and recursively call this function
+            # Note that we only care about inputs (i.e., produced by a VirtualInputInstruction)
+            regs = []
+            if hasattr(producer.inst, "args_in"):
+                regs += producer.inst.args_in
```
hanno-becker, line 794: Why nodes do not have `args_in`/`args_in_out`?

slothy/core/core.py
```diff
@@ -855,7 +855,7 @@ def selftest(self, log):
                 addr = getattr(t.inst, "addr", None)
                 if addr is None:
                     continue
-                addresses.add(addr)
+                addresses = addresses.union(tree.find_all_predecessors_input_registers(t, addr))
```
hanno-becker, line 858: Blocking: This uses the final `reg_state` -- it should use the `reg_state` at the time of `t`.

slothy/core/dataflow.py
```diff
@@ -775,6 +775,30 @@ def has_symbolic_registers(self):
...
+            if hasattr(producer.inst, "args_in_out"):
+                regs += producer.inst.args_in_out
+            predecessors = set()
+            for reg in regs:
+                predecessors = predecessors.union(self.find_all_predecessors_input_registers(producer, reg))
```
hanno-becker, line 799: Why are we forming unions here? In the example
```
vmov rX, sY
ldr tZ, [rX]
```
this would lead to `rX` being marked as an address register, but it shouldn't?

slothy/helper.py
```diff
@@ -1314,8 +1314,8 @@ def run_code(code, txt=None):
                 # If we expect a function return, put a valid address in the LR
                 # that serves as the marker to terminate emulation
                 mu.reg_write(config.arch.RegisterType.unicorn_link_register(), CODE_END)
-            # Setup stack
-            mu.reg_write(config.arch.RegisterType.unicorn_stack_pointer(), STACK_TOP)
+            # Setup stack and allocate allocate initial stack memory
```
hanno-becker, line 1317: typo: allocate allocate

hanno-becker, line 1317:
```suggestion
            # Setup stack and allocate initial stack memory
```

slothy/helper.py
hanno-becker: Pushed a commit addressing this  https://github.com/slothy-optimizer/slothy/pull/125/commits/4313b3de2d66b438bcd15ea4f901c8c036ca0d46

slothy/helper.py
hanno-becker: Fixed by force push

slothy/core/dataflow.py
mkannwischer: it's more that instructions can have either of the two.

PR #128, opened 2024-12-18
Function renaming: Rename .size directives as well

mkannwischer:


dop-amin (approved):
Thanks for adding this.

PR #129, opened 2024-12-18
Armv7-M: Add BranchLoop type

mkannwischer:


slothy/targets/arm_v7m/arch_v7m.py
```diff
@@ -297,6 +297,97 @@ def end(self, other, indentation=0):
         yield f'{indent}cmp {other["cnt"]}, {other["end"]}'
         yield f'{indent}bne {lbl_start}'
 
+
+class BranchLoop(Loop):
```
dop-amin, line 301:
```suggestion
class BranchLoop(Loop):
    """
    More general loop type that just considers the branch instruction as part of the boundary.
    This can help to improve performance as the instructions that belong to handling the loop can be considered by SLOTHY aswell. 
    
    Note: This loop type is still rather experimental. It has a lot of logics inside as it needs to be able to "understand" a variety of different ways to express loops, e.g., how counters get incremented, how registers marking the end of the loop need to be modified in case of software pipelining etc. Currently, this type covers the three other types we offer above, namely `SubsLoop`, `CmpLoop`, and `VmovCmpLoop`. 

    For examples, we refer to the classes `SubsLoop`, `CmpLoop`, and `VmovCmpLoop`. 
 """
```

slothy/core/slothy.py
```diff
@@ -472,7 +472,10 @@ def optimize_loop(self, loop_lbl, postamble_label=None, forced_loop_type=None):
 
         early, body, late, _, other_data, loop = \
             self.arch.Loop.extract(self.source, loop_lbl, forced_loop_type=forced_loop_type)
-        loop_cnt = other_data['cnt']
+        try:
```
dop-amin, line 475: The same `try`-`except` needs to go in `fusion_loop` as in https://github.com/slothy-optimizer/slothy/blob/ce74da9dfc3d3d104665a6849643b1ff60f8cb81/slothy/core/slothy.py#L453. This was removed in fb6fe0db8db40afd0ca2c51c5dc8aedd6bf8ce22.

dop-amin (changes requested):
All good apart from the one fix that needs to be done in `slothy.py` and if you'd like you can accept the comment on the loop class.

dop-amin (approved):

PR #131, opened 2024-12-19
Cortex-M7: Fix pointwise_769_dilithium

mkannwischer:
I made some changes to `pointwise_769_dilithium.s` yesterday when debugging some performance degradation. 
Turns out these changes were not needed and broke the code in some way. reverting my changes and commiting the newly optimized function.


dop-amin:
Test passes on my end. Looks good.

PR #132, opened 2024-12-19
Fix counter correction for SW pipelining with early + late insts

hanno-becker:
When both early and late instructions are used, the loop counter needs adjusting by 2. A recent change broke this because of an unfortunate type confusion leading to the computation of `2 * '1' == '11'` rather than `2 * 1 == 2`.

This commit fixes the computation of the loop counter adjustment.


mkannwischer (approved):
LGTM

PR #135, opened 2024-12-20
Allow boolean SLOTHY tags within macros

mkannwischer:
Previously, a boolean tag (like `// @slothy:core=True`) inside a macro would lead to failure upon trying to unfold macro arguments.

This commit fixes this.

Hoisted out from armv7m branch.


slothy/helper.py
```diff
@@ -783,6 +783,8 @@ def prepare_value(a):
             return a
 
         def apply_arg(l, arg, val):
+            if isinstance(l, str) is False:
```
hanno-becker, line 786:
```suggestion
            # This function is also called on the values of tags, which may not be strings.
            if isinstance(l, str) is False:
```

PR #143, opened 2025-01-03
Extend Spilling 

dop-amin:
Commits extracted from `armv7m` branch as the example for the 769 Dilithium NTT breaks without this.

### DIFF
--- a/slothy/core/config.py
+++ b/slothy/core/config.py
@@ -1005,6 +1005,27 @@ class Config(NestedPrint, LockAttributes):
             return self._allow_spills
 
         @property
+        def spill_type(self):
+            """The type of spills to generate
+
+            This is usually spilling to the stack, but other options may exist.
+            For example, on Armv7-M microcontrollers it can be useful to spill
+            from the GPR file to the FPR file.
+
+            It is expected that this option is set as a dictionary, for example,
+            with the key determining whether the spills are supposed to be to
+            the stack or to the FPR file, and the value defining a starting
+            index for the FPRs in the latter case.
+
+            The exact influence of this option is architecture dependent. You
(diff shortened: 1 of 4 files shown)


slothy/core/config.py
```diff
@@ -1004,6 +1004,19 @@ def allow_spills(self):
...
 
+        @property
+        def spill_type(self):
+            """The type of spills to generate
+
+            This is usually spilling to the stack, but other options may exist.
+            For example, on Armv7-M microcontrollers it can be useful to spill
+            from the GPR file to the FPR file.
+
+            The type of this configuration option is architecture dependent.
```
hanno-becker, line 1015: I forgot the details here... are we always expecting this to be a dictionary of named values to be passed to the members of the `Spill` class?

If so, we could at least modify the getter to return `{}` if `self._spill_type is None`, removing the need for the `if .. else` below.

dop-amin: If I'm not mistaken, the dictionary that is passed here is supposed to be the `**kwargs` in the `Spill` class in the arch model. 

We currently don't use this but in the past we had an example set `slothy.config.constraints.spill_type = { 'spill_to_vreg': 24 }`, which defines the starting point of the FPRs to spill to as an integer. So yes, having a dictionary here makes sense. 

Thus, I think the solution of returning `{}` if `self._spill_type` is `None`, which will pass no key word argument `spill_to_vreg` to `Spill.spill()`, and therefore cause a spill to the stack instead of the FPR is fine.

hanno-becker: @dop-amin Can you document that one always expects a dictionary?

dop-amin: Done :+1:

hanno-becker (approved):

PR #145, opened 2025-01-06
M7: Fix MAC slot constraint + remove (unused) str slot constrainted

mkannwischer:


dop-amin:
Looks good.
Missing that shouldn've had a huge impact.

PR #147, opened 2025-01-06
M7: fix kyber matacc - reserve fpu registers

mkannwischer:
Before the SLOTHY configuration was missing marking the FPU registers as reserved.
Due to that SLOTHY would pick other registers than the ones on the original code.
If those happen to be callee-save registers and are not saved by the function, it won't work.
This was the case for the Kyber mataccs.


dop-amin:
LGTM

PR #148, opened 2025-01-06
Cortex-M7 Kyber Part 2: NTT functions

dop-amin:
This PR adds Kyber's NTT functions. The forward NTT switched to the `Arch_Armv7M.BranchLoop` type in order to get rid of the spilling feature, which allowed the `vmov`+`cmp` loop without `boundaryReserverdRegs`. 

Possible improvments:

- Make address fixup work with the current code.
    - Problem 1: Split heuristic cannot handle an instruction not being found by `id` when enforcing a `before` constraint. Could likely be resolved by allowing the constraint to be optional such that in case the instruction with a certain `id` can't be found AND we're using splitting heuristic, we continue instead of failiing. 
    - Problem 2: Sometimes, instructions still get reordered after the branch which completely breaks the code. Needs investigation. 
- Run with more liberal parameters.
- Also use `Arch_Armv7M.BranchLoop` in the inverse NTT.


dop-amin:
The first point in the list should now be addressed by #150 and #149. 
I re-optimized the code using those changes and (partially) enabling the fixup while switching both examples to the Branch loop type. 

This PR now depends on #150 and #149.

mkannwischer (approved):

PR #149, opened 2025-01-06
Locked ordering: Allow id find to fail with splitting

dop-amin:
In examples where we need to use the `before=my_id` tag together with the splitting heuristic, the optimization would fail because the instruction tagged with `my_id` is not guranteed to appear within the "window" the heuristic is considering at some point. However, in this situation, the constraint can just be omitted as there is no risk of the instructions being placed in an "illegal" way, based on the `before` tag anyways.


mkannwischer (approved):
LGTM. @hanno-becker - any concerns?

PR #150, opened 2025-01-06
Make branch also last instruction without sw pipelining

dop-amin:
In a previous PR (#139), we allowed the branch instruction to be considered as part of the loop body and forced the branch to always be the last instruction in the loop. 
However, we only applied this change for the case where the loop uses software pipelining. This PR enforces this for *all* loops, no matter whether software pipelining is enabled or not.


mkannwischer (approved):
Your comment was misaligned - I fixed it. 
LGTM.

PR #152, opened 2025-01-09
Add four x25519 functions and related assembly classes

SH1E0r1r2y:
Description:
This Pull Request adds four new x25519 functions for M7 and includes the corresponding assembly classes based on these functions. Additionally, it optimizes load instructions (e.g., ldm r8, {r8, r9, r10}) to improve instruction execution order and alignment.


mkannwischer:
This isn't ready yet - we should only merge the x25519 code once it is fully optimized. No point in merging that now already. 

We can, however, start merging the arch+uarch changes. 
As a first start please open a new PR with just the changes you made to `ldrd_imm_splitting_cb` and `ldm_interval_splitting_cb`fixing the splitting of `ldrd` and `ldm` when the address register and output register overlap.

PR #157, opened 2025-01-13
Armv7-M: Default to .w for better alignment

mkannwischer:
Currently, we maintain the instruction-width modifiers as they are in the input code (except for some exceptions). However, this can negatively impact performance as SLOTHY may break code-alignment.
This commit changes the Armv7-M instruction writer, to output .w for all instructions resulting in the best performance (modulo the size of the instruction cache).
Unfortunately, LLVM (in the selftest) stumbles over some of these .w modifiers in places where they do not have any effect. To work around that, we remove the modifiers for the selftest.

### DIFF
--- a/slothy/helper.py
+++ b/slothy/helper.py
@@ -1212,11 +1212,27 @@ class LLVM_Mc():
                                               include=include_paths)
             except subprocess.CalledProcessError as exc:
                 log.error("CPreprocessor failed on the following input")
-                log.error(SouceLine.write_multiline(source))
+                log.error(SourceLine.write_multiline(source))
                 raise LLVM_Mc_Error from exc
 
         if platform.system() == "Darwin":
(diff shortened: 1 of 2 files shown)


slothy/helper.py
```diff
@@ -1212,12 +1212,28 @@ def assemble(source, arch, attr, log, symbol=None, preprocessor=None, include_pa
...
+
+        # Remove all width information - LLVM cannot handle .w for
+        # some instructions that only have a 32-bit encoding,
+        # e.g., uadd16.w works in gcc, but not LLVM.
+        # Unfortunately, for some instructions this depends
+        # on the registers used and, hence, adjusting the input to
+        # SLOTHY is not sufficient.
+        # As currently, we don't have a model of the instruction encodings,
+        # there is no principled way to reason about it.
+        if thumb:
```
dop-amin, line 1230: Probably Arch specific code should not go here. But also don't have a good quick fix in mind right now.

PR #166, opened 2025-03-04
Fix aarch64 loop labels being overwritten

### LINKED ISSUE #165: SLOTHY is dropping textual labels in generated output
I am running SLOTHY on the mlkem-native AArch64 back-end, using a Python 3.11 venv.

See the "make_slothy" branch of the mlkem-native repo.

Slothy produces output OK, but it has replaced textual labels with numeric labels. For example,
in poly_tomont_asm.S the label "poly_tomont_asm_loop" in the "clean" input sources is replaced
by a label "1" in the optimized output.

mlkem-native's "simpasm" script then fails to work on that output.

See attached files for diff of reference vs new optimized code.

[examples.tar.gz](https://github.com/user-attachments/files/19071092/examples.tar.gz)

dop-amin:
This PR addresses the issue #165. In the `SubsLoop` class, the wrong parameter was used in the writer of loop. 

The parameters `lbl_start`, `lbl_end` are some legacy carry-over from the Armv8-M models. I feel like we could remove them globally at some point as they do not add too much value in my opinion. 

* Resolves #165

### DIFF
--- a/examples/opt/aarch64/aarch64_loop_subs_opt_a55.s
+++ b/examples/opt/aarch64/aarch64_loop_subs_opt_a55.s
@@ -1,0 +1,26 @@
+count .req x2
+
+mov count, #16
+start:
+                   // Instructions:    1
+                   // Expected cycles: 1
+                   // Expected IPC:    1.00
+                   //
+                   // Cycle bound:     1.0
+                   // IPC bound:       1.00
+                   //
+                   // Wall time:     0.01s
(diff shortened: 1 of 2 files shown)


hanno-becker:
> The parameters lbl_start, lbl_end are some legacy carry-over from the Armv8-M models. I feel like we could remove them globally at some point as they do not add too much value in my opinion.

Can they at least be removed from the AArch64 model, then?

Issue #164, opened 2025-03-03
Consider NEON ST3 instruction support


rod-chapman:
The mlkem-native Arch64 back-end uses the NEON "st3" instruction in the implementation of the poly_tobytes() function.

This is not currently supported by SLOTHY. An attempt to run SLOTHLY on that code fails with
```
INFO:slothy-cli:Optimizing loop poly_tobytes_asm_asm_loop_start (20 instructions) ...
ERROR:root:Failed to parse instruction st3 {v2.8b, v3.8b, v4.8b}, [x0], #24
```

Could this be considered for enhancement please?

hanno-becker:
@mkannwischer Do you have time to do this?

rod-chapman:
It appears that we have support for ST2 and ST4 ... is ST3 that much harder?

mkannwischer:
> [@mkannwischer](https://github.com/mkannwischer) Do you have time to do this?

Okay - I'll look into it later this week. 

> It appears that we have support for ST2 and ST4 ... is ST3 that much harder?

Of course not. Our models are mostly built lazily, so there are still a lot of instructions that we do not support.

PR #167, opened 2025-03-05
Add ld3+st3 instructions

mkannwischer:
Resolves https://github.com/slothy-optimizer/slothy/issues/164

This adds the ld3 and st3 instructions. 
Regarding the uarch model:
- A55: I followed the A55 SWOG, but this is the worst case performance. There is already a todo there `# TODO: Add distinction between Q/D and B/H vs. D/S`. In case we see poor performance, we may want to revisit this. 
- A72: The current ld2/ld4/st2/st4 models already deviate from what's written in the SWOG. I guessed it somewhere in between.
- Apple Silicon: Similar guesswork...


slothy/targets/aarch64/aarch64_neon.py
```diff
@@ -3014,6 +3014,37 @@ def make(cls, src):
...
+    inputs = ["Va", "Vb", "Vc"]
+    in_outs = ["Xc"]
+    @classmethod
+    def make(cls, src):
+        obj = AArch64Instruction.build(cls, src)
+        obj.addr = obj.args_in_out[0]
+        obj.increment = obj.immediate
+        obj.pre_index = None
+        obj.args_in_combinations = [
+                ( [0,1,2], [ [ f"v{i}", f"v{i+1}", f"v{i+2}" ] for i in range(0,29) ] )
```
hanno-becker, line 3044: Should this be `range(0,30)`? Similar Q for pre-existing models or 2/4-fold ld/st.

mkannwischer: you are right... they are all off by one... 
let me fix that.

mkannwischer: fixed.

hanno-becker: Thanks!

hanno-becker (approved):

PR #172, opened 2025-03-06
Update OR-Tools to v9.12

dop-amin:

PR #175, opened 2025-03-10
OR-Tools v9.12: Use fixed intervals

dop-amin:
This PR is based on #172.

According to a [comment by Laurent Perron](https://github.com/google/or-tools/issues/4166#issuecomment-2503430633), this is supposed to improve efficiency of solving the constraint problem. 

No thorough investigation about efficiency gains in our case has been done.

### DIFF
--- a/.github/workflows/test_basic.yaml
+++ b/.github/workflows/test_basic.yaml
@@ -82,7 +82,7 @@ jobs:
     - uses: ./.github/actions/setup-ubuntu
     - name: Run examples
       run: |
-        python3 example.py --examples ntt_kyber_1_23_45_67_m55,ntt_dilithium_12_34_56_78_m55 --timeout=300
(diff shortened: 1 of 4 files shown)


mkannwischer:
@dop-amin, are you planning to investigate the performance impact of this, or what are we going to do with this PR?

mkannwischer:
> I think using this feature of "newer" ortools versions is the right thing to do as it gives more information to the solver about the structure of the problem. I would think there is almost no way this could hurt the performance. I was waiting to investigate further until we've settled on a version of ortools back then. I don't have the time to follow up on the testing anymore.
> 
> Maybe it's enough to rebase this and roughly check CI times before merging.

Sounds like a good plan to me. Can you rebase it, please?

dop-amin:
I figured this feature seems to be new since 9.10(?) so merging now with 9.7 is incompatible. 
We could leave this as a draft until we update the default or-tools version eventually or close for now and have an issue about this. I think keeping separate paths in the code for either would be a terrible thing to do.

PR #184, opened 2025-04-09
Make aarch64 loop branch instruction flexible

### LINKED ISSUE #181: AArch64: Add support for further loop forms
@nebeid needs support for AArch64 loops with:
* Counter modification: `subs cnt, cnt, #2`
* Branch condition: `b.gt`

dop-amin:
* Resolves #181.

The counter modificaiton by 2 should've been possible before.
I added the branch-instruction as a piece of data that is collected during parsing and emitted accordingly. Note that we need to differentiate for branch types that perform the comparison themselves (e.g., `bnz`) on a register input or branch instructions like `b.gt` relying on flags to be set already. 

The new example also uses software pipelining to demonstrate that the fixup also respects the counter increment of 2.


slothy/targets/aarch64/aarch64_neon.py
```diff
@@ -375,7 +375,7 @@ def __init__(self, lbl="lbl") -> None:
         self.lbl_regex = r"^\s*(?P<label>\w+)\s*:(?P<remainder>.*)$"
         self.end_regex = (
             r"^\s*sub[s]?\s+(?P<cnt>\w+),\s*(?P<reg1>\w+),\s*#(?P<imm>\d+)",
```
hanno-becker, line 377: Nit: This would currently match `subs ... cbnz` which compares twice. Sadly, there still seem to be cases where we are using this pattern, even in mlkem-native.

Can we remember and retain the shape of the `sub[s]` instruction?

dop-amin: I think this was not the case before because of the conditional statement in line 415. However, I now changed it to simply retain the input shape.

dop-amin:
Now depends on #185 to incorporate CI fixes.

PR #185, opened 2025-04-10
Fix selftest for code with aes instructions

### LINKED ISSUE #183: Selftest failing for code using AES instructions
@nebeid observed the selftest failing for code involving `aese` instruction. Probably an architecture flag is missing.

mkannwischer:
This sets the LLVM-MC flags to `aes` such that the selftest can handle aes instructions.
Not all CPUs support aes instructions, but the selftest isn't meant to catch that anyway, so I think setting it always is reasonable. 

Resolves #183


mkannwischer:
There were two problems with the linting.
pydoclint  has a newer version (0.6.5) with more warnings:
```
./slothy/core/dataflow.py:203:1: DOC502 Method `ComputationNode.__init__` has a "Raises" section in the docstring, but there are not "raise" statements in the body
./slothy/core/heuristics.py:321:1: DOC[50](https://github.com/slothy-optimizer/slothy/actions/runs/14370532640/job/40292638281#step:5:51)2 Method `Heuristics.periodic` has a "Raises" section in the docstring, but there are not "raise" statements in the body
./slothy/core/heuristics.py:430:1: DOC503 Method `Heuristics.linear` exceptions in the "Raises" section in the docstring do not match those in the function body. Raised exceptions in the docstring: ['AssertionError', 'SlothyException']. Raised exceptions in the body: ['SlothyException'].
./slothy/core/slothy.py:310:1: DOC502 Method `Slothy.optimize` has a "Raises" section in the docstring, but there are not "raise" statements in the body
./slothy/core/slothy.py:498:1: DOC502 Method `Slothy.fusion_region` has a "Raises" section in the docstring, but there are not "raise" statements in the body
./slothy/core/slothy.py:[52](https://github.com/slothy-optimizer/slothy/actions/runs/14370532640/job/40292638281#step:5:53)6:1: DOC502 Method `Slothy.fusion_loop` has a "Raises" section in the docstring, but there are not "raise" statements in the body
```
I work around this by fixing it to the version we were using before (0.6.2).

I opened https://github.com/slothy-optimizer/slothy/issues/186 so we don't forget too fix this at some point.

The other problem is that the default runners recently switched to Python3.12 by default. 
That leads to other linting warnings:
```
./examples/misc/gen_roots.py:438:63: E231 missing whitespace after ','
./slothy/core/dataflow.py:316:56: E231 missing whitespace after ','
./slothy/core/dataflow.py:31[9](https://github.com/slothy-optimizer/slothy/actions/runs/14371157451/job/40294338053#step:4:10):55: E231 missing whitespace after ','
./slothy/core/dataflow.py:322:59: E231 missing whitespace after ','
```

Actually upgrading Python leads to even bigger problems as the ortools version we are currently using is not compatible with Python 3.12 - so this definitely requires more work. 
I opened https://github.com/slothy-optimizer/slothy/issues/187 for the linting issues and https://github.com/slothy-optimizer/slothy/issues/188 for real support of python3.12.

PR #197, opened 2025-04-10
Remove pandas

mkannwischer:


mkannwischer:
ortool v9.7 incorrectly did not specify pandas as a dependency.
This is fixed in newer ortools versions. After we update to ortools v9.12, we can remove the pandas requirement.

Issue #100, opened 2024-11-06
Re-structure examples


mkannwischer:
I'd like to restructure the examples in SLOTHY. 

I propose the following: 
- naive consists of 1 dir per arch with potentially sub-dirs for different groups (e.g., we may want to group all code that is related to Kyber). 
- opt consits of 1 dir per arch with a subdir per uarch with subdir for the groups mentioned above.

The examples.py gets split up and the actual code gets moved to `naive/<arch>/<group>/example.py` or similar.  

So we have something like 
- example.py
- examples
  - naive
     - aarch64
       - kyber
          -  example.py 
          - xxx.s
     - armv7m
     - armv8m
  - opt
     - aarch64
       - a55
         - kyber
            - xxx_opt_a55.s
         - ....
         
Any comments?
I propose to wait with this restructuring until #61 has been merged.

mkannwischer:
implemented by #211

PR #199, opened 2025-04-22
refactor: Major example refactoring

dop-amin:
This PR addresses #100 and changes the directory structure of the examples directory. It splits the examples by arch and contains one folder for each class of examples (e.g., Kyber-related asm files). In addition, these subfolders also contain the python scripts implementing the optimization for these asm files. example.py on the top-level is thus merely a wrapper for these.

Moreover, the test-cases that check arch-specific operations, e.g., related to parsing have been removed from the examples dir and moved to tests. They can be run using the test.py file.

The class for running an example or a test now resides in common/OptimizationRunner.py.

To add an example/test, perform the following steps:
1. Create a subdir for the arch if not already present inside examples/tests.
2. For examples: Create a subdir for the group of examples. For tests, this is not used yet but could be.
3. Place asm files in the resp. naive directory.
4. Place _example.py/_test.py in the resp. naive directory. See existing files for examples.
5. In top-level example.py/test.py import the list of examples/tests provided by your new script and add it to the list containing all examples/tests.


mkannwischer:
Thanks @dop-amin for all this work! This was annoying me for quite some time already.

> 4. Place _example.py/_test.py in the resp. naive directory. See existing files for examples.

What is the underscore needed for?

dop-amin:
> What is the underscore needed for?
I wanted this file to always appear at the top of the sub directories and not mixed in between the asm files. But we could change that, I don't have a strong preference. 

> Have you looked into how much changes are required in pqax/pqmx, yet?
I have not looked at this yet, but I'd suppose this is mainly about updating the symlinks and maybe adjusting the build system to be able to deal with asm files in sub directories inside the asm folder (in case that's not supported yet, I am not sure). I will take a look at this later.

mkannwischer:
> > What is the underscore needed for?
> 
> I wanted this file to always appear at the top of the sub directories and not mixed in between the asm files. But we could change that, I don't have a strong preference.

That makes sense. Thanks. 


> 
> > Have you looked into how much changes are required in pqax/pqmx, yet?
> 
> I have not looked at this yet, but I'd suppose this is mainly about updating the symlinks and maybe adjusting the build system to be able to deal with asm files in sub directories inside the asm folder (in case that's not supported yet, I am not sure). I will take a look at this later.

I'd prefer if we can merge that at the same time. 
How about we drop the symlinks altogether and just use the path in slothy? With the new build system that should be quite easy.

hanno-becker:
@dop-amin Thank you very much for working on this.

PR #200, opened 2025-04-23
Switch to self-hosted runner

dop-amin:


dop-amin:
> @dop-amin Where is the new runner hosted? What are your plans for this PR?

The runner is hosted locally at MPI-SP. 

My plans are to get things to work. However, I'm experiencing issues with the runner randomly quitting. Running the examples manually works fine. Have not been able to figure out what goes wrong here.

PR #203, opened 2025-04-30
Only reference API-relevant args in `vldr_gather.write()`

### LINKED ISSUE #202: Slothy is not creating isomorphic graphs when performing gather loads.
As described in https://github.com/slothy-optimizer/pqmx/issues/55, there is an issue with Slothy generating incorrect DFGs on Cortex-M55 when using gather loads. The following snippet generates faulty output.

```
slothy_start:
    vldrb.u32 q7, [r0, #0]
    vldrw.u32 q0, [r1, q7]
    vldrb.u32 q6, [r0, #4]
    vldrw.u32 q1, [r1, q6]
    vadd.u32 q2, q1, q0
    vstrw.32 q2, [r2]
slothy_end:
```

This appears to be a bug in register renaming:

```
DEBUG:non-isomorphic-dfg_m55.slothy_start.slothy:Output 0 of 'vldrw.u32 q5, [r1, q7]' renamed to q5
DEBUG:non-isomorphic-dfg_m55.slothy_start.slothy:Input 0 of 'vldrw.u32 q5, [r1, q7]' renamed to r1
DEBUG:non-isomorphic-dfg_m55.slothy_start.slothy:Input 1 of 'vldrw.u32 q5, [r1, q7]' renamed to q1
DEBUG:non-isomorphic-dfg_m55.slothy_start.slothy:New instruction: vldrw.u32 q5, [r1, q7]
```

bremoran:
This change fixes a bug with `vldr_gather` where registers are not renamed correctly. The bug is caused by `vldr_gather.write()` tracking old data from parse-time, rather than updated registers from renaming.

Fixes #202

### DIFF
--- a/slothy/targets/arm_v81m/arch_v81m.py
+++ b/slothy/targets/arm_v81m/arch_v81m.py
@@ -1749,7 +1749,7 @@ class vldr_gather(Instruction):
         uxtw = ""
         if self.uxtw is not None:
             uxtw = f", UXTW #{self.uxtw}"
-        addr = f"[{self.addrgpr}, {self.addrvec}{uxtw}]"
+        addr = f"[{self.args_in[0]}, {self.args_in[1]}{uxtw}]"
 
         return f"vldr{self.width}.{self.datatype} {self.args_out[0]}, {addr}"
 


hanno-becker (commented):
Thank you @bremoran for catching this! Things are more principled in the AArch64 model, but the Armv8.1-M model is unfortunately still quite 'raw'. 

Did you check if there are other instances where the `write()` call refers to parse-time copies of the registers?

bremoran:
I went through every write in the v8.1-M model and I didn't identify any additional instances. I believe that is the only one.

PR #206, opened 2025-05-11
AArch64: Don't enforce whitespacing before/after separating tokens

### LINKED ISSUE #201: Don't treat spaces in AArch64 instruction patterns as mandatory
`aese v0.16b,v1.16b` is not recognized because the spacing in the instruction pattern is mandatory. The pattern substrings ` ,` and `, ` should be replaced by `\s*,` and `,\s*`, respectively (for all instructions, not just `aese`)

hanno-becker:
* Resolves #201 

Previously, a whitespace before/after a separating token such as ',' inside an instruction pattern would be interpreted as mandatory, leading to parsing failures for valid instructions strings such as 'ldr q0,[x0]'.

This commit fixes this by allowing any amount of whitespaces before/after tokens such as ',', '.', '[', ']', regardless of the number of spaces in the instruction pattern.

### DIFF
--- a/examples/naive/aarch64/aarch64_simple0.s
+++ b/examples/naive/aarch64/aarch64_simple0.s
@@ -1,15 +1,15 @@
 ldr q0, [x1, #0]
 ldr q1, [x2, #0]
 
-ldr q8,  [x0]
-ldr q9,  [x0, #1*16]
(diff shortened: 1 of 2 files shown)


hanno-becker:
cc @nebeid

Issue #209, opened 2025-05-13
`b.gt` erroneously requiring a register


nebeid:
I am looking at this small loop; it is recognised by SLOThY only if I (erroneously) put `w6` after `b.gt`

Input
```
	aese	v6.16b,v0.16b
	aesmc	v6.16b,v6.16b
	ld1	{v0.4s},[x4],#16
	aese	v6.16b,v1.16b
	aesmc	v6.16b,v6.16b
	ld1	{v1.4s},[x4],#16
	subs	w6,w6,#2
	b.gt	w6,Loop_enc_iv_enc
```
Output
```
                                      // Instructions:    5
                                      // Expected cycles: 5
                                      // Expected IPC:    1.00
                                      //
                                      // Cycle bound:     5.0
                                      // IPC bound:       1.00
                                      //
                                      // Wall time:     0.01s
                                      // User time:     0.01s
                                      //
                                      // ----- cycle (expected) ------>
                                      // 0                        25
                                      // |------------------------|----
        aese v6.16b, v0.16b           // *.............................
        ld1 {v0.4S}, [x4], #16        // *.............................
        aesmc v6.16b, v6.16b          // ..*...........................
        aese v6.16b, v1.16b           // ....*.........................
        ld1 {v1.4S}, [x4], #16        // ....*.........................

                                       // ------ cycle (expected) ------>
                                       // 0                        25
                                       // |------------------------|-----
        // aese v6.16b, v0.16b         // *..............................
        // aesmc v6.16b, v6.16b        // ..*............................
        // aese v6.16b, v1.16b         // ....*..........................
        // ld1 {v0.4S}, [x4], #16      // *..............................
        // ld1 {v1.4S}, [x4], #16      // ....*..........................

        sub w6, w6, #2
Loop_enc_iv_enc:
                                      // Instructions:    6
                                      // Expected cycles: 8
                                      // Expected IPC:    0.75
                                      //
                                      // Cycle bound:     8.0
                                      // IPC bound:       0.75
                                      //
                                      // Wall time:     0.02s
                                      // User time:     0.02s
                                      //
                                      // ----- cycle (expected) ------>
                                      // 0                        25
                                      // |------------------------|----
        aesmc v6.16b, v6.16b          // l.............................
        aese v6.16b, v0.16b           // ..*...........................
        aesmc v6.16b, v6.16b          // ....*.........................
        aese v6.16b, v1.16b           // ......*.......................
[...]

PR #212, opened 2025-05-21
Fix b.gt loop

dop-amin:
Addresses #209. 

* Split parsing for loops that have a register in the branch instructions, and the ones that don't
* Adjust respective example to contain valid asm


mkannwischer (approved):
Looks good to me. Thanks @dop-amin. 

@nebeid can you confirm this solves your problem?

mkannwischer (changes requested):
After discussing with @dop-amin, I think this needs more work. 

1) b.gt and bgt are the same instruction. Both should be supported. 
2) All other conditions (ne, lt, etc.) should also be supported
3) cbnz does not exist without a register. 
4) bnz does not exist at all
5) bne does not exist with a register.

nebeid:
> @nebeid can you confirm this solves your problem?
I don't find it working on my end. It's still requiring a register with `b.gt`(I don't think it's my setup).

dop-amin:
> After discussing with @dop-amin, I think this needs more work.
> 
>     1. b.gt and bgt are the same instruction. Both should be supported.
> 
>     2. All other conditions (ne, lt, etc.) should also be supported
> 
>     3. cbnz does not exist without a register.
> 
>     4. bnz does not exist at all
> 
>     5. bne does not exist with a register.

I've addressed all of these.

dop-amin:
> > @nebeid can you confirm this solves your problem?
> 
> I don't find it working on my end. It's still requiring a register with `b.gt`(I don't think it's my setup).

I've tried your example from #209 and it works for me without a register for `b.gt`. Can you send exactly what you've tried so I can debug?

nebeid:
> > > @nebeid can you confirm this solves your problem?
> > 
> > 
> > I don't find it working on my end. It's still requiring a register with `b.gt`(I don't think it's my setup).
> 
> I've tried your example from #209 and it works for me without a register for `b.gt`. Can you send exactly what you've tried so I can debug?

Taking that back. I tried the PR (hadn't noticed it wasn't merged yet) and it worked with my small example. Thank you. Will try `b.hs` and will report back if I find an issue.

mkannwischer (approved):
LGTM. Thanks @dop-amin!

PR #213, opened 2025-05-21
Switch to self-hosted linux runner

dop-amin:
Try if a self-hosted linux runner works, in constrast to #200.


dop-amin:
This was mostly for testing purposes.

PR #216, opened 2025-05-22
AArch64: Relax spacing around immediates, add some shifted logical instructions

hanno-becker:
* Based on #212 

Found during some joint debugging with @nebeid


mkannwischer (approved):
LGTM

Issue #215, opened 2025-05-22
Cleanly separate between examples, functional tests, and regression test


dop-amin:
#199 and #211 already separate examples and functional tests more cleanly.

However, to detect performance regressions within SLOTHY (or its dependencies), we would like to be able to get proper "benchmarks" on its performance. Currently, we're mainly looking at the full NTT optimization runs in CI to do this. However, this approach is not very principled, as it is subject to random variances based on how "lucky" the solver gets. 

I would propose a few things:

- Run a set of representative examples with full optimizations in CI (probably depending on #200)
- Run each of those examples multiple times to factor out the solver's luck
- Output the performance metrics in a structured format to make them easily trackable
- Use these metrics to detect regressions inside new PRs

PR #218, opened 2025-05-23
Regression

dop-amin:
Adds a simple framework for running regression tetsts on SLOTHY's performance to CI.
Partially addresses #215.


dop-amin:
For some reason `ntt_dilithium_123_45678_a55` keeps failing due to immediate out of range issues. This was not the case with the previous optimization run in CI. Needs investigation.

regression.py
```diff
@@ -0,0 +1,157 @@
...
+            log_model=args.log_model,
+            debug_logfile=args.debug_logfile,
+            log_model_dir=args.log_model_dir,
+            timeout=args.timeout,
+            only_target=args.only_target,
+        )
+        reg_res[n] = t
+
+    summary_file = os.environ.get("GITHUB_STEP_SUMMARY")
+    with open(summary_file, "w") as f:
```
mkannwischer, line 148: can you wrap this in an 
`if sumary_file is not None:` 
please? otherwise this errors outside of CI.

PR #222, opened 2025-05-26
AArch64: Fix register constraints for 2-register st1

hanno-becker:
Unlike `stp`, the input registers must be consecutive.


mkannwischer:
`encoded as "Rt" plus 1 modulo 32.` sounds like there can be a wrap-around just like for st2/3/4?

hanno-becker:
Yes, though let's do that all at once -- #217 is open for that.

mkannwischer (approved):

PR #225, opened 2025-05-28
Achieve compatibility with Python 3.9-3.13 and add tests to CI

mkannwischer:


mkannwischer:
Giving up on Python 3.8 - it's EOL since Oct 2024 anyway. There's at least one more problem with it - `removeprefix`.

dop-amin:
Looks good to me, 23mins for neon NTT and 2mins for helium NTT is also the expected time for it to take. 
Thanks!

PR #232, opened 2025-06-02
refactor: remove the unuse instruction, vhcsub, vcaddf, vcsubf

### LINKED ISSUE #229: MVE: Remove `vhcsub`, `vcsubf`
I have no idea what the `vhcsub`, `vcsubf` instructions are that are currently present in our MVE model.
They are never used in our code. 

We should remove them from both the architecture model as well as the microarchitectual models of the M55 and M85.

willieyz:
- Remove the unuse functions "vhcsub", "vcaddf", "vcsubf" from slothy
- Resolve #229


willieyz:
remove the vhcsub, vcsubf, and the vcaddf is the vcadd for the floating-point data type, we leave it alone just for now.

PR #234, opened 2025-06-03
Refactor MVE architecture model

mkannwischer:
Our first model was the Armv8.1-M model and it contained a lot of ad-hoc
parsing code. When writing the Neon model, we came up with a much cleaner
way of describing instructions and abstracting away most of the parsing code.
This commit aligns the old Armv8.1-M model with the new way of describing
instructions.

Refactoring was primarily done by @willieyz! Thank you!

A full optimization run for M55 and M85 was performed and the files were tested in pqm4: https://github.com/slothy-optimizer/pqmx/pull/65. However, we do not include the re-optimized file in this PR as they have no real value. A couple of opt files were missing, however. Those were added.

There are some follow-up issues identified in the process that we'll resolve later:
 - https://github.com/slothy-optimizer/slothy/issues/235 
 - https://github.com/slothy-optimizer/slothy/issues/243
 - https://github.com/slothy-optimizer/slothy/issues/246


.python-version
mkannwischer, line 1: please remove this file

examples/misc/gen_roots.py
```diff
@@ -435,7 +435,7 @@ def get_roots_of_unity(self):
             if self.print_label:
                 yield (
                     f"roots_l"
-                    f"{''.join([str(i) for i in range(cur_iter,cur_iter+merged)])}:"
+                    f"{''.join([str(i) for i in range(cur_iter, cur_iter+merged)])}:"
```
mkannwischer: this should not be changed

slothy/core/dataflow.py
mkannwischer, line 1: please do not change this file

slothy/targets/arm_v81m/arch_v81m.py
```diff
@@ -37,7 +37,8 @@
 import re
 import math
 
-from sympy import simplify
+# from sympy import simplify
```
mkannwischer: remove the comment

tests/naive/armv8m/instructions.o
mkannwischer, line 1: please remove this file

PR #241, opened 2025-06-12
Add Neoverse N1 model for ST3 instruction.

rod-chapman:
Add Neoverse N1 model for ST3 instruction.

This is used in the poly_tobytes() function in mlkem-native.

Instruction latency and throughput data from the Neoverse N1 Software Optimization Guide, table 34, p. 55.

Note that poly_tobytes() using the "multiple structures, Q form, storing bytes" form of ST3, so we choose these data from table 34.


mkannwischer (approved):
LGTM. I double-checked the SWOG.

Issue #242, opened 2025-06-16
Armv8-M: Add some missing instructions


mkannwischer:
We require a few more instruction to be supported by the Armv8-M model.

- vmov
- vmov (double to double)
- vmov (2 registers to double)
- vmsr
- vsub (vector - GPR)
- vstr (scatter store)

@willieyz, could you please add them?

PR #244, opened 2025-06-16
Support vmov and more

willieyz:
Solves #242 242
We require a few more instruction to be supported by the Armv8-M model.

- [ ] - vmov
- [ ] - vmov (double to double)
- [ ] - vmov (2 registers to double)
- [ ] - vmsr
- [ ] - vsub (vector - GPR)
- [ ] - vstr (scatter store)


slothy/targets/arm_v81m/arch_v81m.py
```diff
@@ -2250,6 +2278,12 @@ def make(cls, src):
         return obj
 
 
+class vmsr(MVEInstruction):
+    pattern = "vmsr <Rs>, <Rn>"
```
mkannwischer: This doesn't do what you think it does. Rs will be seen as a symbolic register and replaced by an `r` register, so that doesn't work. 
You will have to hardcode the possible flags. i.e., use 
```
pattern = "vmsr FPSCR, <Rn>"
```

But maybe we can avoid using `vmsr` altogether. let me check .

slothy/targets/arm_v81m/arch_v81m.py
```diff
@@ -1493,6 +1501,20 @@ def make(cls, src):
         return obj
 
 
+class vstr_scatter_stored(MVEInstruction):
```
mkannwischer: Please align the naming with `vldrw`, i.e., `vstrw_scatter`.

slothy/targets/arm_v81m/arch_v81m.py
```diff
@@ -1493,6 +1501,20 @@ def make(cls, src):
...
+    pattern = "vstrw.<dt> <Qd>, [<Rn>, <Qm>, UXTW <imm>]"
```
mkannwischer, line 1520: The UXTW should be optional, please add both variants.
Do they exist with byte and halfword datatypes as well? If yes, please add them.

slothy/targets/arm_v81m/arch_v81m.py
```diff
@@ -1493,6 +1501,20 @@ def make(cls, src):
...
+    inputs = ["Qm", "Rn"]
+    outputs = ["Qd"]
```
mkannwischer: `Qd` is not an output. This is a store instruction. Everything is an input.

mkannwischer (changes requested):
Could you please change the Armv8-M instruction test to 

```
class Instructions(OptimizationRunner):
    def __init__(self):
        super().__init__("instructions", base_dir="tests")

    def core(self, slothy):
        slothy.config.allow_useless_instructions = True
        slothy.optimize(start="start", end="end")
```
i.e., allow reordering and disable functional only optimization.
If you do that you see some errors. Please fix them.

slothy/targets/arm_v81m/cortex_m55r1.py
```diff
@@ -539,6 +541,7 @@ def get_min_max_objective(slothy):
         vmulf_T2,
         vfma,
     ): 2,
+    vmsr_FPSCR: 4,
```
mkannwischer: This is not what I see in the M55 SWOG. It says latency 1 and throughput 1. 
Where do you get the 4 from?

mkannwischer (changes requested):

mkannwischer (approved):
LGTM. Thanks @willieyz

Issue #235, opened 2025-06-06
Remove typing hints


mkannwischer:
In the original MVE model it was possible to use symbolic registers without specifiying their register type (q vector registers vs r GP registers). This can lead to SLOTHY not being able to pick the right instruction when parsing. To solve this, for MVE typing_hints were introduced as an configuration option allowing the globally set the type of a symbolic registers. 
For AArcch64 this approach was insufficient due to the register views (e.g., w0 being the lower half of x0). Instead we introduced typing annotations for symbolic register, e.g., `w<tmp>` would be a symbolic w register. This way slothy can track that actually `x<tmp>` and `w<tmp>` are actually the same value and still SLOTHY can correctly parse the right instruction variant.

Currently the MVE model is being refactored to follow the same approach as for AArch64. 
Once that is completed we should entirely remove the typing hints from the SLOTHY configuration and all examples.

PR #248, opened 2025-06-17
Remove typing hints configuration option

willieyz:
- Solves #235 

Since we introduced typing annotations for symbolic register, e.g., r<tmp> and refactor the MVE model to follow the same approach as for AArch64, we should entirely remove the typing hints from the SLOTHY configuration and all examples.


mkannwischer (approved):
Thanks @willieyz. LGTM.

Issue #243, opened 2025-06-16
Armv8-M: Allow disabling `unsafe_address_offset_fixup`


mkannwischer:
Currently we enforce unsafe_address_offset_fixup to be enabled for Armv8-M - see https://github.com/slothy-optimizer/slothy/blob/915c224166207ce07b31152194305c3b6687d09b/slothy/core/config.py#L1490.
After we have merged #234, we can allow disabling it.
The only change that is required for that is remodelling the address registers in incrementing memory operations as in/out instead of just input.

PR #255, opened 2025-06-27
Allow disabling unsafe_address_offset_fixup for Armv8-M

mkannwischer:
This PR implements issue #243 by enabling the ability to disable `unsafe_address_offset_fixup` for ARM v8.1-M architecture targets.

### DIFF
--- a/slothy/core/config.py
+++ b/slothy/core/config.py
@@ -251,12 +251,6 @@ class Config(NestedPrint, LockAttributes):
 
         .. note::
 
-            For historical reason, this feature cannot be disabled for
-            the Armv8.1-M architecture model. A refactoring of that model is needed
-            to make address offset fixup configurable.
-
-        .. note::
-
             The user-imposed safety constraint is not a necessity -- in principle,
             SLOTHY could detect when it is safe to reorder ldr/str instructions with
             increment.
@@ -1485,8 +1479,6 @@ class Config(NestedPrint, LockAttributes):
 
     @unsafe_address_offset_fixup.setter
     def unsafe_address_offset_fixup(self, val):
-        if val is False and self.arch.arch_name == "Arm_v81M":
-            raise InvalidConfig("unsafe address offset fixup must be set for Armv8.1-M")
         self._unsafe_address_offset_fixup = val
 
     @locked_registers.setter
--- a/slothy/targets/arm_v81m/arch_v81m.py
+++ b/slothy/targets/arm_v81m/arch_v81m.py
@@ -1109,29 +1109,29 @@ class ldrd_no_imm(MVEInstruction):
 
 class ldrd_with_writeback(MVEInstruction):
     pattern = "ldrd <Rt0>, <Rt1>, [<Rn>, <imm>]!"
-    inputs = ["Rn"]
     outputs = ["Rt0", "Rt1"]
+    in_outs = ["Rn"]
 
     @classmethod
     def make(cls, src):
         obj = MVEInstruction.build(cls, src)
         obj.increment = obj.immediate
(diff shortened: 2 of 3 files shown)


mkannwischer:
Somewhat scary. Amazon Q took 30 minutes for this with these instructions: 
```
Your job is to implement #243. That requires two changes: First you need to change slothy/targets/arm_v81m/arch_v81m.py to change all memory operations (ld/st) with increment such that the address register is marked as an in_out rather than just an input. Note that pre-indexed variants (i.e., those that don't inrement the address) do not need to be updated. Second you need to adapt the config.py such that offset fixup can be disabled also for Arm_v81M. Also update the documentation. To test the feature change the tests/naive/armv8m/_test.py such that the Example0 has slothy.config.unsafe_address_offset_fixup = False set. Once you are done make sure all tests (python3 test.py) run successfully. Finally make sure that python3 example.py --dry run runs through successfully for the M55 and M85 (you can use the --target input to limit it to just these cores). Once you are done create acommit in a new branch, push it to Github and open a draft pull request. Acknowledge in the PR title and and description that you have done this work and include this promt in the PR description. For fun also include the duration it took you to implement this feature (now until the time you open the PR). 
```

tests/naive/armv8m/_test.py
```diff
@@ -45,6 +45,11 @@ class Example0(OptimizationRunner):
     def __init__(self):
         super().__init__("simple0", base_dir="tests")
 
+    def core(self, slothy):
+        slothy.config.unsafe_address_offset_fixup = False
+        slothy.config.outputs = ["inA", "inB"]
```
mkannwischer: this looks wrong. you should be able to remove this.

slothy/targets/arm_v81m/arch_v81m.py
```diff
@@ -2099,62 +2116,62 @@ def make(cls, src):
...
         obj.increment = 64
 
         return obj
 
 
 class vst42_with_writeback(MVEInstruction):
     pattern = "vst42.<dt> {<Qd0>, <Qd1>, <Qd2>, <Qd3>}, [<Rn>]!"
-    inputs = ["Rn"]
-    in_outs = ["Qd0", "Qd1", "Qd2", "Qd3"]
+    inputs = []
```
mkannwischer: remove the empty input initializations. they are not needed.

slothy/core/config.py
```diff
@@ -251,9 +251,11 @@ def unsafe_address_offset_fixup(self):
...
         .. note::
 
-            For historical reason, this feature cannot be disabled for
-            the Armv8.1-M architecture model. A refactoring of that model is needed
-            to make address offset fixup configurable.
+            This feature is now configurable for all supported architectures,
+            including Armv8.1-M. When disabled for Armv8.1-M, memory operations
+            with address register increment (writeback and post-increment variants)
+            will have their address registers modeled as input-output rather than
+            just input, allowing for more accurate dependency tracking.
```
mkannwischer: remove this.

mkannwischer:
Thank you for the review! I've addressed all the comments:

1. **Removed empty inputs initializations**: Cleaned up all the  lines from the architecture file as they are not needed when using .

2. **Fixed test configuration**: Changed from manually declaring  to using  which is the proper way to handle this case.

3. **Simplified documentation**: Removed the detailed explanation from the config.py comment as requested.

All tests continue to pass successfully after these changes. The implementation correctly demonstrates that with , the optimization requires 1 stall instead of 0, showing the more conservative but accurate instruction scheduling behavior.

The changes are now cleaner and follow the established patterns in the codebase.

PR #260, opened 2025-07-01
Armv8-M: Add missing hint support (required for memory annotations)

mkannwischer:
Needed for #256


dop-amin:
I believe what you implemented looks sufficient for the "simple" case to work. In aarch64, we also have a pattern inside instructions that supports taking a hint as an argument, e.g., `x_ldr_stack_imm_with_hint`. However, I don't recall how that works and it could be added when needed at some later point.

Issue #265, opened 2025-07-02
Request adding `q_ldp_with_postinc` instruction to Neoverse N1


pennyannn:
When optimizing an assembly program with instruction `ldp q0, q1, [x0], #0x50`, I got the following error:
```
slothy.targets.common.UnknownInstruction: Couldn't find <class 'slothy.targets.aarch64.aarch64_neon.q_ldp_with_postinc'> for ldp q0, q1, [x0], #0x50
```

Could we import this instruction into Neoverse N1?

mkannwischer:
It has been added. Does that solve your problem?

There are a few more instructions missing (see #267) - we'll try to add them as well but that may take a little longer. If you urgently need an instruction in the meantime, please don't hestitate to open another issue.

PR #266, opened 2025-07-03
Add Ldp_Q to Neoverse N1 model

mkannwischer:
This adds Ldp_Q (immediate offset or post-increment) to the Neoverse N1 model. From the [SWOG](https://developer.arm.com/documentation/109896/latest/) (page 33):
 - latency 7
 - throughput 1
 - used pipelines: L and I (However, we simplify this to only L as for the other instructions).

Resolves https://github.com/slothy-optimizer/slothy/issues/265

### DIFF
--- a/slothy/targets/aarch64/neoverse_n1_experimental.py
+++ b/slothy/targets/aarch64/neoverse_n1_experimental.py
@@ -43,6 +43,7 @@ from slothy.targets.aarch64.aarch64_neon import (
     Ldr_Q,
     Str_Q,
     Stp_Q,
+    Ldp_Q,
     Vrev,
     uaddlp,
     vmov,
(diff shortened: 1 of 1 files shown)


slothy/targets/aarch64/neoverse_n1_experimental.py
```diff
@@ -242,6 +243,7 @@ def get_min_max_objective(slothy):
 }
 
 default_latencies = {
+    Ldp_Q: 7,
```
hanno-becker: For OOO uArch's we tend to be a bit more relaxed about the latency modelling of load instructions since it is expected that the uArch will heavily front-load them anyway. I would suggest to stick with 4 here and leave a comment.

mkannwischer: Thanks. Fixed.

hanno-becker (approved):

PR #269, opened 2025-07-05
Add `AESInstruction` to A55 and A72 uArch models

mkannwischer:

PR #270, opened 2025-07-05
AArch64: Add `vext` to A55+A72, add `vbic`, `vand`, `veor` to A72. 

### LINKED ISSUE #268: `pmull`/`pmull2`/`ext` missing from A55 uArch model
I'm running on input
```asm
ldr q0, [x1, #0]
ldr q1, [x2, #0]
pmull v2.1q, v0.1d, v1.1d
pmull2 v3.1q, v0.2d, v1.2d
ext v4.16b, v0.16b, v0.16b, #8
ext v5.16b, v1.16b, v1.16b, #8
eor v4.16b, v0.16b, v4.16b
eor v5.16b, v1.16b, v5.16b
pmull v4.1q, v4.1d, v5.1d
eor v29.16b, v29.16b, v2.16b
eor v30.16b, v30.16b, v3.16b
eor v31.16b, v31.16b, v4.16b
st1 {v29.16b}, [x1], #0
st1 {v30.16b}, [x2], #0
st1 {v31.16b}, [x3], #0
```
and the last line of the output is ``slothy.targets.common.UnknownInstruction: Couldn't find <class 'slothy.targets.aarch64.aarch64_neon.pmull1_q'> for pmull v2.1q, v0.1d, v1.1d``.
Yet it appears [here](https://github.com/slothy-optimizer/slothy/blob/0d5b46c8cf29bfc7ef7c3a4d70417f1299e5edbc/slothy/targets/aarch64/aarch64_neon.py#L3439) this is exactly how `pmull` is expected to be written.
If I replace `pmull` with `mul` I get the same error on `ext` (following `pmull/pmull2`)

mkannwischer:
- Depends on #269
- Fixes #268


slothy/targets/aarch64/cortex_a55.py
```diff
@@ -451,13 +458,15 @@ def get_min_max_objective(slothy):
     (vsri): 1,
     (vusra): 1,
     (vand, vbic, veor): 1,
+    vext: 1,
```
dop-amin, line 461: One question about this: In the SWOG footnote 1 on `vext` it says "If the instruction has Q-form, the Q-form of the instruction can only be dual issued as instruction 0 and execution throughput is 1." 
Do you think this is something we should consider? Possibly add a slot constraint? 

I would be fine merging this without addressing it as we don't seem to take any special action on this for, e.g., `trn1` which has exactly the same characteristics as `vext`.

mkannwischer: This is taken care of by 
```
    is_qform_form_of(vext): [[ExecutionUnit.VEC0, ExecutionUnit.VEC1]],
    is_dform_form_of(vext): [ExecutionUnit.VEC0, ExecutionUnit.VEC1],
 ```

mkannwischer: Or well, partially. I don't think we ever model the slot constraint, do we?

dop-amin: True, did not recall it also solves the issue about the throughput implicitly. Then, everything's fine for me!

dop-amin: > Or well, partially. I don't think we ever model the slot constraint, do we?

I think we've tried but deemed it as non-useful as it blew up complexity for little use.

PR #271, opened 2025-07-06
AArch64: Clean-up logical Neon instructions and shift+insert Neon instructions

mkannwischer:
- Depends on #270

See individual commit messages.


slothy/targets/aarch64/neoverse_n1_experimental.py
```diff
@@ -183,11 +184,20 @@ def get_min_max_objective(slothy):
     (vmovi): ExecutionUnit.V(),
     (vand, vadd, vsub): ExecutionUnit.V(),
     (vxtn): ExecutionUnit.V(),
-    (vuxtl, vshl, vshl_d, vshli, vsrshr, vshrn, vsshr): ExecutionUnit.V1(),
+    (
+        vuxtl,
+        vshl,
+        vshl_d,
+        AArch64NeonShiftInsert,
```
dop-amin, line 191: More of a general comment: It's very subtle in this case that the super class would shadow a class listed explicitly; also because it's not immediately obvious which classes are children to which super class. I've fallen for this in other instances before and it's not great that the order in which the elements are listed determines which match is picked. 

Maybe a convetion of adding the super classes last could limit the impact of this? We could also address this in the lookup where we'd not just pick the first, but the most specific match.

dop-amin (approved):
All good, my comment it just general discussion.

slothy/targets/aarch64/neoverse_n1_experimental.py
mkannwischer: What do you think about erroring in case there is more than one match? It seems very likely that it is a mistake in the model if that happens.

dop-amin: Would also work. I'll open an issue.

PR #277, opened 2025-07-16
Resolve Issue #226 - Saving the example's config

### LINKED ISSUE #226: Dump SLOTHY config alongside/inside optimized asm file
Sometimes, it's hard to keep track of which optimization parameters and setup has been used to produce a certain snippet of assembly.
It may be usefeul to have a feature that dumps the SLOTHY config/python script steps next to the optimized file or even into it.

thisisjube:
While executing the example, write the configuration to a file in the same location as the optimized code to track the settings.

Closes #226

### DIFF
--- a/common/OptimizationRunner.py
+++ b/common/OptimizationRunner.py
@@ -241,7 +241,8 @@ class OptimizationRunner:
         example_name = self.__class__.__name__
         subfolder_full = arch_label_dict[self.arch] + "/" + self.subfolder
         example_config_path = f"{self.base_dir}/naive/{subfolder_full}_example.py"
-        out_name = f"{self.base_dir}/opt/{subfolder_full}{self.outfile}.conf"
+        outfile = self.outfile.split("/")[-1].split(".")[0]
+        out_name = f"{self.base_dir}/opt/{subfolder_full}{outfile}.conf"
         self.extract_class_source(example_config_path, example_name, out_name)
 
     def extract_class_source(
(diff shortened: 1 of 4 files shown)


mkannwischer:
Thanks @thisisjube for looking into this!
Some comments: 
 - Can this be a draft please until you have completed it, please?
 - Note that SLOTHY can also be used using the slothy-cli in which case you won't have the Python configuration file. I think it would be better to rely on the effective configuration when optimize() is called and dump that rather than reading other files.

thisisjube:
> Thanks @thisisjube for looking into this! Some comments:
> 
> * Can this be a draft please until you have completed it, please?
> * Note that SLOTHY can also be used using the slothy-cli in which case you won't have the Python configuration file. I think it would be better to rely on the effective configuration when optimize() is called and dump that rather than reading other files.

- Yes sure, this was actually meant to be a draft
- Fair point. I will discuss details with Amin :)

mkannwischer:
> > Thanks @thisisjube for looking into this! Some comments:
> > ```
> > * Note that SLOTHY can also be used using the slothy-cli in which case you won't have the Python configuration file. I think it would be better to rely on the effective configuration when optimize() is called and dump that rather than reading other files.
> > ```
> 
> I think it would be good if we could have one, common interface between the Python examples and the CLI. @mkannwischer What do you think about making the CLI "transfer" the config into an `OptimizationRunner` instance internally?

What is wrong with just dumping the properties of the Config object (maybe ignoring those that have default values) at each call to optimize/optimize_loop?

PR #279, opened 2025-07-17
AArch64: Various instructions needed for ML-KEM rejection sampling

willieyz:
This PR add instruction support for the rej_uniform_loop48 in mlkem-native
- https://github.com/pq-code-package/mlkem-native/issues/1117

The following is the missing instructions list:

- `bic tmp0.8h, #0xf0, lsl 8`
  - A55 SWOG (page 34)
  - A72 SWOG (page 24)
  - N1 SWOG (page 37)

- `cmhi sign0.8h, mlkem_q.8h, val0.8h`
  - A55 SWOG (page 34)
  - A72 SWOG (page 24)
  - N1 SWOG (page 37)

- `ldr q24, [x3, x12, lsl #4]`
  - A55 SWOG (page 30)
  - A72 SWOG (page 13)
  - N1 SWOG (page 20)

- `cnt sign0.16b, sign0.16b`
  - A55 SWOG (page 38)
  - A72 SWOG (page 29)
  - N1 SWOG (page 47)

- `tbl val0.16b, {val0.16b}, table0.16b`
  - A55 SWOG (page 39)
  - A72 SWOG (page 30)
  - N1 SWOG (page 48)

- `sub buflen buflen #48`
  - A55 SWOG (page 34)
  - A72 SWOG (page 24)
  - N1 SWOG (page 37)

Execution unit, latency and throughput data come from:
1: a55: https://developer.arm.com/documentation/epm128372/latest/
2: a72: https://developer.arm.com/documentation/uan0016/latest/
3: neoverse_n1: https://developer.arm.com/documentation/109896/latest/

PR #283, opened 2025-07-22
AArch64: add `s` register support and add `fmov` and `uaddlv` instructions

mkannwischer:
This PR
- [X] Extends the AArch64 model to support the s register aliases
- [X] Adds `uaddlv` support to the AArch64 model and A55/A72/N1 uArch models
- [x] Adds  `fmov` support to the AArch64 model and A55/A72/N1 uArch models
- Related to #279

### DIFF
--- a/slothy/targets/aarch64/aarch64_neon.py
+++ b/slothy/targets/aarch64/aarch64_neon.py
@@ -834,7 +834,7 @@ class AArch64Instruction(Instruction):
                 f"<(?P<symbol_{g.group(1)}{g.group(2)}>\\w+)>))"
             )
 
-        src = re.sub(r"<([BHWXVQTD])(\w+)>", pattern_transform, src)
+        src = re.sub(r"<([BHWXVQTDS])(\w+)>", pattern_transform, src)
 
         # Replace <key> or <key0>, <key1>, ... with pattern
         def replace_placeholders(src, mnemonic_key, regexp, group_name):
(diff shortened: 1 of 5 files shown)


slothy/targets/aarch64/aarch64_neon.py
```diff
@@ -3585,6 +3585,12 @@ class Fmov(AArch64Instruction):
     pass
 
 
+class fmov_s_form(Fmov):
+    pattern = "fmov <Wd>, <Sa>"
+    inputs = ["Sa"]
+    in_outs = ["Wd"]
```
mkannwischer: Why would this be an in_out?

willieyz: No, this shouldn't
sorry for this mistake, fixed now, thanks you for your reviewing!

slothy/targets/aarch64/cortex_a55.py
```diff
@@ -261,6 +262,7 @@ def get_min_max_objective(slothy):
         AArch64NeonLogical,
         sub_imm,
         vuaddlv_sform,
+        fmov_s_form,  # from double/single to gen reg
```
mkannwischer: Something is off here. 
You write `can issue from both slot`, but you mark it as occupying both vector units, i.e., no dual issuing. 
Looking into the SWOG, I think you comment is correct, but you code is wrong. It should be allowed to dual issue.

willieyz: Got it, fixed.

mkannwischer (commented):
Thanks @willieyz. LGTM now.

PR #288, opened 2025-07-25
Add support for non-power-of-2 unroll factors and support unrolling with unknown iteration count. 

### LINKED ISSUE #286: Add support for loop unrolling + unknown_iteration_count
When using loop unrolling, we currrently assume that the number of iterations is a multiple of the unrolling factor.

SLOTHY should support the combination of unrolling and an unknown iteration count. The pattern to be produced should look like this:

```
if loop_cnt < UNROLL:
  jump TAIL

loop:
   BODY
   sub loop_cnt, loop_cnt, UNROLL
   cmp loop_cnt, UNROLL
   branch_if_greater_equal

TAIL:
   cmp loop_cnt #1
   jump TAIL_1
  
  cmp loop_cnt #2 
  jump TAIL_2

  ...
```

When SW pipelining is used, this also needs to take into account the number of exceptional iterations, e.g. for a loop with early but no late instructions:

```
cmp loop_cnt, UNROLL:
branch_if_smaller TAIL

... PREAMBLE FOR UNROLLED LOOP ...
sub loop_cnt, loop_cnt, UNROLL
cmp loop_cnt, UNROLL
branch_if_smaller POSTAMBLE

LOOP:
   BODY
   sub loop_cnt, loop_cnt, UNROLL
   cmp loop_cnt, UNROLL
   branch_if_greater_equal LOOP

POSTAMBLE:
 ... POSTAMBLE FOR UNROLLED LOOP...

TAIL:
   cmp loop_cnt #1
   jump TAIL_1
  
  cmp loop_cnt #2 
  jump TAIL_2

  ...
```

mkannwischer:
- Resolves https://github.com/slothy-optimizer/slothy/issues/286

Disclaimer: This was primarily written by Claude Code with some help from Amazon Q. I have done quite a few iterations, but still care should be taken verifying the changes made.

- Add Branch class methods for handling division and remainder operations
  - Power-of-2: Use LSR/AND instructions
  - Non-power-of-2: Use UDIV/MLS instructions with temporary registers

- Add test cases for both power-of-2 (unroll=4) and non-power-of-2 (unroll=3)
  - AArch64: aarch64_unknown_iter{,_pow2}.s
  - ARMv7-M: test_unknown_iter{,_pow2}.s

- Update loop classes to handle arbitrary unroll factors
  - Remove power-of-2 assertions
  - Add temp_reg_for_division parameter
  - Use Branch.divide_by_constant for division

### DIFF
--- a/slothy/core/slothy.py
+++ b/slothy/core/slothy.py
@@ -657,7 +657,7 @@ class Slothy:
             current_reserved.add(remainder_reg)
             c.reserved_regs = current_reserved
             self._tail_remainder_reg = remainder_reg
-            self._tail_divisor_reg = available_gprs[1]
+            self._unrolling_trailing_tmp_reg = available_gprs[1]
 
         preamble_code, kernel_code, postamble_code, num_exceptional = (
             Heuristics.periodic(body, logger, c)
@@ -718,6 +718,9 @@ class Slothy:
 
         optimized_code += indented(preamble_code)
 
+        # Initialize unrolling_trailing_tmp_reg to None by default
+        unrolling_trailing_tmp_reg = None
+
         if self.config.sw_pipelining.unknown_iteration_count:
             if postamble_label is None:
                 postamble_label = f"{loop_lbl}_postamble"
@@ -726,7 +729,7 @@ class Slothy:
 
                 # Use pre-allocated registers (allocated before kernel optimization)
                 temp_reg = self._tail_remainder_reg
-                divisor_reg = self._tail_divisor_reg
+                unrolling_trailing_tmp_reg = self._unrolling_trailing_tmp_reg
 
                 optimized_code += indented(
                     self.arch.Branch.if_less_than(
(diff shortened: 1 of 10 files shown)


slothy/core/slothy.py
```diff
@@ -708,7 +755,48 @@ def loop_lbl_iter(i):
         optimized_code += indented(postamble_code)
 
         if self.config.sw_pipelining.unknown_iteration_count:
+            # Add jump to tail section for remainder handling after main loop
+            if self.config.sw_pipelining.unroll > 1:
+                tail_lbl = f"{loop_lbl}_tail"
+                temp_reg = self._tail_remainder_reg
```
mkannwischer, line 786: Something is off here. How can one be sure that the tail_remainder_reg has not been used in the main loop body? If it is used, it breaks.

mkannwischer: I marked it as reserved now.

PR #290, opened 2025-07-31
Feat: Supporting immediate values without #

### LINKED ISSUE #284: AArch64: Consider supporting immediate values without #
In the process of applying SLOTHY to ML-KEM rejection sampling, we noticed that 
```
bic     v0.8h, #0xf0, lsl 8`
```
cannot be parsed by SLOTHY as it expects
```
bic     v0.8h, #0xf0, lsl #8`
```

That is likely a much broader limitation. Maybe we can consider making the parsing of immediates more flexible.

willieyz:
- Resolves #284 
- This commit suppoting immediate values without `#`
- Following model are supported:
  - `aarch64`
  - `arch_v7m`
  - `arch_v81m`
1. The `imm_pattern` was updated to recognize immediate values regardless of the presence of `#`, and the `<imm>` prefix strip off logic was adjusted accordingly.
2. Additionally, commas `,` were removed from `imm_pattern` to avoid misinterpreting operands like `q2, UXTW 2` as a single immediate:
```
.syntax unified
start:
vstrw.32 q0, [r0, #16]
vstrw.u32 q0, [r0, q2, UXTW 2].
end:
```
3. To support symbolic aliases like `#(SHIFT)`, used in instructions such as follow, we now combine the old symbolic-style and the new numeric literal patterns. This allows both formats to be parsed properly.
```
vrshr.s32 q0, q1, #(SHIFT)
```

### DIFF
--- a/slothy/targets/aarch64/aarch64_neon.py
+++ b/slothy/targets/aarch64/aarch64_neon.py
@@ -874,7 +874,11 @@ class AArch64Instruction(Instruction):
 
         flag_pattern = "|".join(flaglist)
         dt_pattern = "(?:|2|4|8|16)(?:B|H|S|D|b|h|s|d)"
-        imm_pattern = "#(\\\\w|\\\\s|/| |-|\\*|\\+|\\(|\\)|=|,)+"
+        imm_pattern = (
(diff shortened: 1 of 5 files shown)


mkannwischer (approved):
Thanks @willieyz. Great & simple improvement that is long overdue.

Issue #294, opened 2025-08-01
Log to stdout by default


mkannwischer:
Currently, when using Python to invoke SLOTHY, it is up to the user to configure logging.
In almost all cases one will do:
```
import logging
logging.basicConfig(stream=sys.stdout, level=logging.INFO)
```

When skipping this configuration step, SLOTHY does not log anything at all.
We should change the default behavior to logging to stdout.

PR #299, opened 2025-08-04
Switch default logging to stdout; default log level INFO

mkannwischer:
Before, SLOTHY would not log anything by default leaving it up to the user to configure logging.
This is tedious and confusing.
This commit changes: If no loggers are configured, it defaults to logging to stdout with log level INFO.
The README and the tutorial are adjusted accoringly.

Resolves https://github.com/slothy-optimizer/slothy/issues/294

### DIFF
--- a/slothy/core/slothy.py
+++ b/slothy/core/slothy.py
@@ -48,6 +48,7 @@ one-shot and heuristic optimizations using SLOTHY.
 """
 
 import logging
+import sys
 from types import SimpleNamespace
 
 from slothy.core.dataflow import DataFlowGraph as DFG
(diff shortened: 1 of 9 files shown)


slothy/core/slothy.py
```diff
@@ -97,7 +98,15 @@ def _get_target(self):
 
     def __init__(self, arch, target, logger=None):
         self.config = Config(arch, target)
-        self.logger = logger if logger is not None else logging.getLogger("slothy")
+
+        # Configure default logging to stdout if no handlers are configured
+        if logger is None:
+            root_logger = logging.getLogger()
+            if not root_logger.handlers:
```
dop-amin, line 105: Just for my understanding: This check is required in case there's some wrapper around SLOTHY which already instantiated an unnamed (root) logger, that is already configured in a certain way, right?

mkannwischer: There are two ways to configure loggers in SLOTHY: 
1) pass the logger argument to the Slothy constructor
2) Globally configure logging using the logging module

The outer conditional catches the first, the inner catches the second. 

At least that is my understanding.

PR #306, opened 2025-08-06
Aarch64:  add conditional compare instructions

cheng-wei-huang0612:
I have added a group of instructions:
Instruction group: Conditional Compare
Instructions: ccmp, ccmn

The parameters are taken from the following references
SWOG A55 page 18
SWOG A72 page 8

I have also add some "xzr-variant" parser definitions, e.g.,

```
class ccmp_xzr2(AArch64ConditionalCompare):
    pattern = "ccmp <Xa>, xzr, <imm>, <flag>"
    inputs = ["Xa"]
    modifiesFlags = True
    dependsOnFlags = True
```

and tests are provided in the instructions.s

I want to mention that I am hesitant to add parameters to m1 coldstorm, since according to the https://dougallj.github.io/applecpu/icestorm-int.html, which indicates that:

CCMN                                      LAT = 1,          TP = 0.556    Uops = 1,        Int = 1    Units = u1-3
CCMP                                      LAT = 1,          TP = 0.556    Uops = 1,        Int = 1    Units = u1-3

where

TP: reciprocal throughput (cycles per instruction when throughput bound)

So, the inverse_throughput should be computed as
0.556 * number_of_units
= 0.556 * 3
= 1.668

which, I think google OR-tool is not going to accept this floating point number input.
(I did not test what will I get if we plug this 1.668 into apple_m1_firestorm_experimental.py as test.py does not contain any test with m1 core)


slothy/targets/aarch64/cortex_a72_frontend.py
```diff
@@ -200,6 +201,7 @@ def get_min_max_objective(slothy):
         ExecutionUnit.ASIMD1,
     ],
     AArch64NeonShiftInsert: [ExecutionUnit.ASIMD1],
+    AArch64ConditionalCompare: [ExecutionUnit.INT()],
```
mkannwischer: Why does this occupy both integer pipelines? Shouldn't it be just one?
```suggestion
    AArch64ConditionalCompare: ExecutionUnit.INT(),
```

mkannwischer (changes requested):
Thanks @cheng-wei-huang0612. 
I think you got the execution units on A72 wrong, but please double check.
Rest looks good!

slothy/targets/aarch64/cortex_a72_frontend.py
cheng-wei-huang0612: According to a72 SWOG, it is indicated that ccmn, ccmp can go to both integer pipelines:

Screenshot 2025-08-08 at 11 19 11 AM

mkannwischer: Okay, but that is not what you modelled, is it? It should be changed from occupying I0 AND I1 to occupying I0 OR I1

cheng-wei-huang0612: So, if I model
```
AArch64ConditionalCompare: [ExecutionUnit.INT()],
```
then SLOTHY treats this as: AArch64ConditionalCompare occupy both I0 AND I1
 
if I model
```
AArch64ConditionalCompare: ExecutionUnit.INT(),
```
then AArch64ConditionalCompare occupy I0 OR I1

mkannwischer: Yes,  remember that ExecutionUnit.INT() resolves to `[ExecutionUnit.INT0, ExecutionUnit.INT1]` that means INT0 OR INT1
`[[ExecutionUnit.INT0, ExecutionUnit.INT1]]` means INT0 AND INT1.

I know this can be confusing and we may have gotten it wrong sometimes.

cheng-wei-huang0612: Ok, that is clear, thanks !!

mkannwischer: Thanks!

mkannwischer (approved):
Thanks @cheng-wei-huang0612. LGTM. 

I took the liberty to rebase on top of main and resolve the conflict.
I also combined your 5 commits into 1.

PR #311, opened 2025-08-13
Fix slothy-cli parser to accept "flags" as register name

### LINKED ISSUE #208: Fix slothy-cli command line parser to accept "flags" as a register name
When using "flags" in the slothy-cli, command line, it is interpreted the homonymous `RegisterType`. This makes it impossible to specify that the flags register is an output when calling SLOTHY through slothy-cli.

mkannwischer:
Remove RegisterType.from_string() conversion that was converting register names like "flags" to enums, causing AttributeError when the rest of the codebase expects strings. Also allow list-to-set type conversion.

Fixes #208


mkannwischer:
@hanno-becker - I am unsure if this is the right fix.
The 
```
       # Try to parse as RegisterType
        ty = arch.RegisterType.from_string(val)
        if ty is not None:
            logger.debug("Value %s parsed as RegisterType", val)
            return ty
```
never seemed to have worked as intended - at least for the outputs.
But I am unsure if this does break anything for any other case and we have very little CI for the slothy-cli.

Issue #68, opened 2024-07-10
AAarch64 architectural model cannot handle barrel shift with different capitalization


mkannwischer:
Works fine: 
```
eor sAga, tmp,  sAga_, ror #39
```
This breaks:
```
eor sAga, tmp,  sAga_, ROR #39
```
```
Traceback (most recent call last):
  File "/home/mjk/git/pqax/slothy/example.py", line 1579, in <module>
    main()
  File "/home/mjk/git/pqax/slothy/example.py", line 1574, in main
    run_example(e, debug=args.debug, dry_run=args.dry_run,
  File "/home/mjk/git/pqax/slothy/example.py", line 1570, in run_example
    ex.run(**kwargs)
  File "/home/mjk/git/pqax/slothy/example.py", line 151, in run
    self.core(slothy, *self.extra_args)
  File "/home/mjk/git/pqax/slothy/example.py", line 1380, in core
    slothy.optimize(start="loop_0", end="end_loop_0")
  File "/home/mjk/git/pqax/slothy/slothy/core/slothy.py", line 259, in optimize
    early, core, late, num_exceptional = Heuristics.periodic(body, logger, c)
                                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/mjk/git/pqax/slothy/slothy/core/heuristics.py", line 302, in periodic
    dfg = DFG(body, logger.getChild("dfg_generate_outputs"),
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/mjk/git/pqax/slothy/slothy/core/dataflow.py", line 583, in __init__
    self.src = self._parse_source(src)
               ^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/mjk/git/pqax/slothy/slothy/core/dataflow.py", line 629, in _parse_source
    return list(map(self._parse_line, SourceLine.reduce_source(src)))
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/mjk/git/pqax/slothy/slothy/core/dataflow.py", line 621, in _parse_line
    insts = self.arch.Instruction.parser(l)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #315, opened 2025-08-20
AArch64: add instruction parsing barrel shift case-insensitive for `ror` mnemonics

willieyz:
- Resolves: #68 

- This PR fixes the case-sensitive problem for the barrel shift instruction.
- We ported `barrel_pattern` from the arch_v7m model to handle barrel shift parsing instead of using hardcoded patterns. Compared with arch_v7m, we added an additional `i` in `barrel_pattern = "(?i:lsl|ror|lsr|asr)\\s*"` to support case-insensitive mnemonics.

> Note: currently, the `<barrel>` pattern is only applied to `eor_ror`. Should we consider applying this method to all instructions that use barrel shifts?


slothy/targets/aarch64/aarch64_neon.py
```diff
@@ -2422,52 +2427,33 @@ class AArch64ShiftedArithmetic(AArch64Instruction):
...
-    modifiesFlags = True
-
-
-class add_asr(AArch64ShiftedArithmetic):
-    pattern = "add <Xd>, <Xa>, <Xb>, asr <imm>"
-    inputs = ["Xa", "Xb"]
-    outputs = ["Xd"]
-
-
 class add_imm_lsl(AArch64ShiftedArithmetic):
```
mkannwischer: should be lsl -> shifted

slothy/targets/aarch64/aarch64_neon.py
```diff
@@ -3173,7 +3147,7 @@ class vbic(AArch64NeonLogical):
 
 
 class vbic_imm_lsl(AArch64NeonLogical):
```
mkannwischer: should be lsl -> shifted

slothy/targets/aarch64/aarch64_neon.py
```diff
@@ -1844,7 +1849,7 @@ def write(self):
 
 
 class x_ldr_with_imm_lsl(Ldr_X):
```
mkannwischer: should be lsl -> shifted.

mkannwischer (approved):
LGTM. Thanks @willieyz!

Issue #219, opened 2025-05-26
AArch64: Don't force numbering of `<dt>` annotations in instruction pattern


hanno-becker:
AArch64 instruction patterns cannot include multiple `.<dt>` specifiers. In this case, they must be numbered as `.<dt0>, .<dt1>, ..`. This is unexpected, undocumented, and no useful error message is given when it happens.

- [ ] At the least, fail with an expressive error message if an instruction pattern includes more than one `.<dt>`
- [ ] Likely better, allow repeated `.<dt>` size specifiers, and force the respective specifiers to be equal during parsing

Noted by @nebeid

PR #316, opened 2025-08-21
AArch64: Add support for repeated .<dt> specifiers

willieyz:
- Resolves: #219

- Add support for using repeated `.<dt>` in instruction patterns, while keeping backward compatibility.
- This PR solve problem by:
  - Introducing a new static method `_replace_duplicate_mnemonicKey()` in `AArch64Instruction` to handle duplicate `<dt>` specifiers during parsing and writing.
  - Enforcing that all `.<dt*>` are equal during parsing.
    - Here is an example:
      - For pattern: `smlsl <Vd>.<dt0>, <Va>.<dt1>, <Vb>.<dt1>`
      - When parsing:
        - `smlsl v6.4s, v7.4h, v8.4h` -> Passed,
        - `smlsl v6.4s, v7.4h, v8.4s` -> Failed,
           return `return "Inconsistent dataname: <dt0> vs <dt1> for datatype: 4s"`
        - `smlsl v6.4s, v7.4h, v8.16b` -> Failed,
           return `"Inconsistent datatype: 4h vs 16b for dataname: <dt1>"`

### DIFF
--- a/slothy/targets/aarch64/aarch64_neon.py
+++ b/slothy/targets/aarch64/aarch64_neon.py
@@ -820,6 +820,20 @@ class AArch64Instruction(Instruction):
         return src
 
     @staticmethod
+    def _enforce_datatype_matching(pattern, res):
+        datatypes = {}
+        for i, m in enumerate(re.finditer(r"<dt\d*>", pattern)):
+            dt = m.group(0)
+            val = res.get(f"datatype{i}", res.get("datatype"))
+            if dt in datatypes and datatypes[dt] != val:
+                raise FatalParsingException(
+                    f"Inconsistent data type: {datatypes[dt]} vs {val}"
+                )
+            elif dt not in datatypes and val in datatypes.values():
+                raise FatalParsingException(f"Inconsistent dt: {dt}")
+            datatypes[dt] = val
+
+    @staticmethod
     def _unfold_pattern(src):
 
         # Those replacements may look pointless, but they replace
@@ -1103,6 +1117,8 @@ class AArch64Instruction(Instruction):
         else:
             assert isinstance(src, dict)
             res = src
+
+        AArch64Instruction._enforce_datatype_matching(pattern, res)
 
         obj = c(
             pattern,
--- a/slothy/targets/aarch64/aarch64_neon.py
+++ b/slothy/targets/aarch64/aarch64_neon.py
@@ -1247,13 +1247,13 @@ class nop(AArch64Instruction):
 
 
 class vadd(AArch64Instruction):
-    pattern = "add <Va>.<dt0>, <Vb>.<dt1>, <Vc>.<dt2>"
+    pattern = "add <Va>.<dt>, <Vb>.<dt>, <Vc>.<dt>"
     inputs = ["Vb", "Vc"]
     outputs = ["Va"]
 
 
 class vsub(AArch64Instruction):
-    pattern = "sub <Va>.<dt0>, <Vb>.<dt1>, <Vc>.<dt2>"
+    pattern = "sub <Va>.<dt>, <Vb>.<dt>, <Vc>.<dt>"
     inputs = ["Vb", "Vc"]
     outputs = ["Va"]
 
@@ -1386,7 +1386,7 @@ class Q_Ld2_Lane_Post_Inc(AArch64Instruction):
 
 
 class q_ld2_lane_post_inc(Q_Ld2_Lane_Post_Inc):
-    pattern = "ld2 { <Va>.<dt0>, <Vb>.<dt1> }[<index>], [<Xa>], <imm>"
+    pattern = "ld2 { <Va>.<dt>, <Vb>.<dt> }[<index>], [<Xa>], <imm>"
     in_outs = ["Va", "Vb", "Xa"]
 
     @classmethod
@@ -1403,7 +1403,7 @@ class q_ld2_lane_post_inc(Q_Ld2_Lane_Post_Inc):
 
 
 class q_ld2_lane_post_inc_force_output(Q_Ld2_Lane_Post_Inc):
-    pattern = "ld2 { <Va>.<dt0>, <Vb>.<dt1> }[<index>], [<Xa>], <imm>"
+    pattern = "ld2 { <Va>.<dt>, <Vb>.<dt> }[<index>], [<Xa>], <imm>"
     # TODO: Model sp dependency
     in_outs = ["Xa"]
     outputs = ["Va", "Vb"]
@@ -2973,25 +2973,25 @@ class Vzip(AArch64Instruction):
 
 
 class vzip1(Vzip):
-    pattern = "zip1 <Vd>.<dt0>, <Va>.<dt1>, <Vb>.<dt2>"
+    pattern = "zip1 <Vd>.<dt>, <Va>.<dt>, <Vb>.<dt>"
     inputs = ["Va", "Vb"]
     outputs = ["Vd"]
 
 
 class vzip2(Vzip):
-    pattern = "zip2 <Vd>.<dt0>, <Va>.<dt1>, <Vb>.<dt2>"
+    pattern = "zip2 <Vd>.<dt>, <Va>.<dt>, <Vb>.<dt>"
     inputs = ["Va", "Vb"]
     outputs = ["Vd"]
 
 
 class vuzp1(Vzip):
-    pattern = "uzp1 <Vd>.<dt0>, <Va>.<dt1>, <Vb>.<dt2>"
+    pattern = "uzp1 <Vd>.<dt>, <Va>.<dt>, <Vb>.<dt>"
     inputs = ["Va", "Vb"]
     outputs = ["Vd"]
 
 
 class vuzp2(Vzip):
-    pattern = "uzp2 <Vd>.<dt0>, <Va>.<dt1>, <Vb>.<dt2>"
+    pattern = "uzp2 <Vd>.<dt>, <Va>.<dt>, <Vb>.<dt>"
     inputs = ["Va", "Vb"]
     outputs = ["Vd"]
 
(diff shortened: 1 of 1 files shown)


slothy/targets/aarch64/aarch64_neon.py
```diff
@@ -3497,7 +3509,7 @@ class vumlsl2(Vmlal):
 
 
 class vsmlsl(Vmlal):
-    pattern = "smlsl <Vd>.<dt0>, <Va>.<dt1>, <Vb>.<dt2>"
+    pattern = "smlsl <Vd>.<dt>, <Va>.<dt>, <Vb>.<dt>"
```
mkannwischer: This does not make sense? 
A valid instruction would be `smlsl v6.4s, v7.4h, v8.4h`, so the 3 datatypes are not equal? 
I would rather have expected 
`pattern = "smlsl <Vd>.<dt0>, <Va>.<dt1>, <Vb>.<dt1>`. 

Also, the original issue mentioned that the equality should be enfored, i.e., `smlsl v6.4s, v7.4s, v8.4h` should fail parsing.Do we actually going to insist on that @hanno-becker? The assembler is going to complain about that anyway.

slothy/targets/aarch64/aarch64_neon.py
```diff
@@ -807,6 +807,17 @@ class AArch64Instruction(Instruction):
...
+    def _replace_duplicate_mnemonicKey(src, mnemonic_key):
+
+        pattern = re.compile(rf"<{re.escape(mnemonic_key)}\d*>")
+
+        matches = pattern.findall(src)
+        if len(matches) > 1:
+            for i, match in enumerate(matches):
+                src = re.sub(re.escape(match), f"<{mnemonic_key}{i}>", src, count=1)
+        return src
+
```
mkannwischer, line 821: Why does this need to be a static functions instead of a local function inside of `_unfold_pattern` just like `replace_placeholders`? I don't see any benefit of doing that.

slothy/targets/aarch64/aarch64_neon.py
```diff
@@ -3471,7 +3484,7 @@ class vumlsl2(Vmlal):
 
 
 class vsmlsl(Vmlal):
-    pattern = "smlsl <Vd>.<dt0>, <Va>.<dt1>, <Vb>.<dt2>"
+    pattern = "smlsl <Vd>.<dt0>, <Va>.<dt1>, <Vb>.<dt1>"
```
mkannwischer, line 3516: @willieyz. Can you please explain what you are doing now? Do I understand correctly that you are supporting repeated `<dt1>`, but are not enforcing that all `<dt1>` are equal on parsing?

willieyz: Hello @mkannwischer, thank you for taking the time to review my change.
I’m currently working on fixing the issue you pointed out. 
At the moment, I have only implemented support for repeated <dt*>, but I haven’t yet added enforcement to ensure that all <dt*> values are equal during parsing, actually, not sure where to start at this part and still looking at it!

slothy/targets/aarch64/aarch64_neon.py
willieyz: Yes, you are right, thank you for pointing this out! 
Moving this function into `_unfold_pattern` as a local function will be a better approach. Thanks again!

slothy/targets/aarch64/aarch64_neon.py
mkannwischer: If this is too much hassle, we may decide to skip that part.
After all, the goal of SLOTHY is not to check that the input assembly is correct.
@hanno-becker, would you be happy of we just write `<dt>` everywhere?

slothy/targets/aarch64/aarch64_neon.py
willieyz: Hello @mkannwischer,
Regarding this issue, after reviewing the code, I recall that I wrote this as a static method because it is not only used in `_unfold_pattern`, but also in the `write()` function.
However, I think your suggestion makes sense. I also don’t like how it currently looks, maybe add it to the `helper.py`?

slothy/targets/aarch64/aarch64_neon.py
willieyz: Hello @mkannwischer, I have added the check to enforce `. <dt*>` datatype equality during the parsing part in a separate commit. 
May I kindly ask if this aligns with what you had in mind?
Also, I feel there must be some other better approach than this one; if you have any suggestions, please feel free to let me know. 
Thank you for your help!

slothy/targets/aarch64/aarch64_neon.py
```diff
@@ -807,6 +807,19 @@ class AArch64Instruction(Instruction):
 
     PARSERS = {}
 
+    @staticmethod
+    def _replace_duplicate_mnemonicKey(src, mnemonic_key):
```
mkannwischer: Can we not mix camel case with underscores please? 
Also what is the rationale behind mnemonicKey? This has nothing to do with the mnemonic?
How about `_replace_duplicate_datatypes`?

slothy/targets/aarch64/aarch64_neon.py
```diff
@@ -903,6 +917,30 @@ def _parse(line):
...
+            # smlsl v6.4s, v7.4h, v8.4h -> Passed
+            # smlsl v6.4s, v7.4h, v8.4s -> Failed,
+            # return "Inconsistent datatype for <dt1>: 4h vs 4s"
+            datatypes = {}
+
+            for s_op, l_op in zip(src.split(","), line.split(",")):
+
+                if "." in s_op.strip() and "." in l_op.strip():
+                    dt_name = s_op.split(".", 1)[1]
+                    dt_value = l_op.split(".", 1)[1]
```
mkannwischer: It is not a good idea to perform datatype parsing yourself. We have already implemented datatype parsing somewhere (search datatype). Please use the parsed value of those please.
The rest of this code is fine.

slothy/targets/aarch64/aarch64_neon.py
mkannwischer: Okay.

slothy/targets/aarch64/aarch64_neon.py
mkannwischer: Okay. let's just leave as is.

slothy/targets/aarch64/aarch64_neon.py
```diff
@@ -3471,7 +3510,7 @@ class vumlsl2(Vmlal):
 
 
 class vsmlsl(Vmlal):
-    pattern = "smlsl <Vd>.<dt0>, <Va>.<dt1>, <Vb>.<dt2>"
+    pattern = "smlsl <Vd>.<dt0>, <Va>.<dt1>, <Vb>.<dt1>"
```
mkannwischer, line 3516: You will still have to make these changes for all instructions.

slothy/targets/aarch64/aarch64_neon.py
willieyz: Hello, @mkannwischer, thank you for your review!
I had add logic to enforcing the consistent between datatype and .<dt*> specifiers,
The detail are explained in PR description, if there is any problem, please let me know, I will fix it as soon as possible, thank you!

slothy/targets/aarch64/aarch64_neon.py
willieyz: Fixed

slothy/targets/aarch64/aarch64_neon.py
willieyz: Fixed.

slothy/targets/aarch64/aarch64_neon.py
willieyz: Fixed, 
now all patterns in `aarch64_neon.py` are aligned with the new rules: enforce consistency between datatype and 
`.<dt>` specifiers, also, allow duplicate datatypes.

slothy/targets/aarch64/aarch64_neon.py
```diff
@@ -1090,6 +1104,45 @@ def build(c, src):
...
+                        f" for datatype: {dt_value}"
+                    )
+            else:
+                if dt_name in datatypes.values():
+                    conflict = next(k for k, v in datatypes.items() if v == dt_name)
+                    raise Instruction.ParsingException(
+                        f"Inconsistent datatype: {conflict} vs {dt_value}"
+                        f" for dataname: {dt_name}"
+                    )
+                datatypes[dt_value] = dt_name
```
mkannwischer: Can you elaborate why you need all that complexity? 
What is wrong with just dong something simple like 
```py
    @staticmethod
    def _enforce_datatype_matching(pattern, res):
        datatypes = {}
        for i, m in enumerate(re.finditer(r"<dt\d*>", pattern)):
            dt = m.group(0)
            val = res.get(f"datatype{i}", res.get("datatype"))
            if dt in datatypes and datatypes[dt] != val:
                raise FatalParsingException(f'Inconsistent data type: {datatypes[dt]} vs {val}')
            datatypes[dt] = val
```

(Also note that this is using a `FatalParsingException` rather than a ParsingException so that we actually see the error instead of just getting output saying it ccould not parse; and that I moved it to a separate function to not pollute the `build` method too much.

willieyz: Hello @mkannwischer, thank you for the review.
This method is indeed simpler and cleaner than my original PR, but should we also add an additional check to ensure that `smlsl v6.4h, v7.4h, v8.4h` also fails?
If so, the final code block would look like this:
```python
    @staticmethod
    def _enforce_datatype_matching(pattern, res):
        datatypes = {}
        for i, m in enumerate(re.finditer(r"<dt\d*>", pattern)):
            dt = m.group(0)
            val = res.get(f"datatype{i}", res.get("datatype"))
            if dt in datatypes and datatypes[dt] != val:
                raise FatalParsingException(f'Inconsistent data type: {datatypes[dt]} vs {val}')
            elif dt not in datatypes and val in datatypes.values():   //additional check
                raise FatalParsingException(f'Inconsistent dt')
            datatypes[dt] = val
```

mkannwischer: sounds good.

willieyz: Fixed

mkannwischer (approved):
LGTM. Thanks @willieyz!

Issue #272, opened 2025-07-07
Fail on multiple matches in uArch model look ups


dop-amin:
As per the discussion in #271, we should probably improve the operation in case there are multiple matches when doing a multi dict lookup inside the uArch model. RIght now, we simply select the first match. In case a super class is shadowing the entry of a children class, this is unnoticed. 

One solution to the problem would be to always pick the most specific match (from a class hierarchy perspective). 
Another solution that would be more explicit from a user's perspective is to simply fail on multiple matches. The only disadvantage would be that overriding one of the properties while following the others of a super class would not be supported.

PR #325, opened 2025-09-12
Fail if microarchitectual model contains multiple classes of an instruction's class hierarchy

willieyz:
- This PR resolves: #272 
- This PR add failed on multiple matches in `lookup_multidict`, and solved the conflict for each multiple matches.
- The following 2 multiple matches worth notice:
  - `vmlaldava`
    - The conflict arises in default_latencies (`vmlaldava: 2` vs. `vmlaldava: 3`).  I resolved it by removing `vmlaldava: 2`, since according to the [M55 SWOG (MVE Integer Vector instructions)](https://developer.arm.com/documentation/102692/0101/Instruction-latencies/MVE-Integer-Vector-instructions), this instruction has two latencies:  3 for first-time use and 2 for repeated use, I think?
    - For now, I applied a latency of 3 universally.  Do we need to implement a mechanism to handle multiple latencies in this scenario?

  - `vhcadd`
    - according to the [M55 SWOG (MVE Integer Vector instructions)](https://developer.arm.com/documentation/102692/0101/Instruction-latencies/MVE-Integer-Vector-instructions), for the multiple matches failure(`vhcadd: 1` v.s. `vhcadd: 2`) I correct the latency to 1, and remove the `vhcadd: 2`

### DIFF
--- a/slothy/targets/aarch64/aarch64_big_experimental.py
+++ b/slothy/targets/aarch64/aarch64_big_experimental.py
@@ -30,9 +30,10 @@
 #
 
 from enum import Enum
+from slothy.helper import lookup_multidict
 from slothy.targets.aarch64.aarch64_neon import (
     is_neon_instruction,
-    lookup_multidict,
+    find_class,
     aese_x4,
     aesr_x4,
     Ldp_X,
@@ -274,12 +275,14 @@ default_latencies = {
 
 
 def get_latency(src, out_idx, dst):
-    latency = lookup_multidict(default_latencies, src)
+    instclass_src = find_class(src)
+    latency = lookup_multidict(default_latencies, src, instclass_src)
     return latency
 
 
 def get_units(src):
-    units = lookup_multidict(execution_units, src)
+    instclass_src = find_class(src)
+    units = lookup_multidict(execution_units, src, instclass_src)
     if isinstance(units, list):
         return units
     else:
@@ -287,4 +290,5 @@ def get_units(src):
 
 
 def get_inverse_throughput(src):
-    return lookup_multidict(inverse_throughput, src)
+    instclass_src = find_class(src)
+    return lookup_multidict(inverse_throughput, src, instclass_src)
--- a/slothy/targets/aarch64/aarch64_neon.py
+++ b/slothy/targets/aarch64/aarch64_neon.py
@@ -38,7 +38,6 @@ similar to those used in the Arm ARM.
 """
 
 import logging
-import inspect
 import re
 import math
 from enum import Enum
@@ -4763,36 +4762,3 @@ def all_subclass_leaves(c):
 
 
 Instruction.all_subclass_leaves = all_subclass_leaves(Instruction)
-
-
-def lookup_multidict(d, inst):
-    instclass = find_class(inst)
-    result = None
-
-    for ll, v in d.items():
-        # Multidict entries can be the following:
-        # - An instruction class. It matches any instruction of that class.
-        # - A callable. It matches any instruction returning `True` when passed
-        #   to the callable.
-        # - A tuple of instruction classes or callables. It matches any instruction
-        #   which matches at least one element in the tuple.
-        def match(x):
-            if inspect.isclass(x):
-                return isinstance(inst, x)
-            assert callable(x)
-            return x(inst)
-
-        if not isinstance(ll, tuple):
-            ll = [ll]
-        for lp in ll:
-            if match(lp):
-                if result is not None:
-                    raise UnknownInstruction(
-                        f"Multiple matches found for {instclass} for {inst}"
-                    )
-                result = v
-                break
-
-    if result is None:
-        raise UnknownInstruction(f"Couldn't find {instclass} for {inst}")
-    return result
--- a/slothy/targets/aarch64/apple_m1_firestorm_experimental.py
+++ b/slothy/targets/aarch64/apple_m1_firestorm_experimental.py
@@ -32,10 +32,9 @@
 
 from enum import Enum
 from itertools import combinations, product
-
+from slothy.helper import lookup_multidict
 
 from slothy.targets.aarch64.aarch64_neon import (
-    lookup_multidict,
     find_class,
     Ldr_X,
     Str_X,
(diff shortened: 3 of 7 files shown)


slothy/targets/aarch64/aarch64_neon.py
```diff
@@ -4810,7 +4812,13 @@ def match(x):
             ll = [ll]
         for lp in ll:
             if match(lp):
-                return v
-    if default is None:
+                if result is not None:
```
mkannwischer: This does not make sense if default is not None.
If I see correctly, we are not using the default argument anywhere. Maybe just remove it?

willieyz: Hello @mkannwischer, thanks for your review!  
You are right, the `default` argument was never used.  
I had removed it.

slothy/helper.py
```diff
@@ -1801,3 +1802,40 @@ def extract(arch: any, source: list, lbl: str, forced_loop_type: any = None) ->
                 pass
 
         raise FatalParsingException(f"Couldn't identify loop {lbl}")
+
+
+class UnknownInstruction(Exception):
+    pass
```
mkannwischer: Re-defining this class here is not a good idea. Can we have a single definition of that class please?

slothy/targets/aarch64/aarch64_big_experimental.py
```diff
@@ -274,17 +275,17 @@ def get_min_max_objective(slothy):
 
 
 def get_latency(src, out_idx, dst):
-    latency = lookup_multidict(default_latencies, src)
+    latency = lookup_multidict(default_latencies, src, find_class)
```
mkannwischer: I don't like that the find_class function is passed as a reference here. Would it be possible to pass the itererator over the instruction classes instead? That is a little cleaner.

slothy/helper.py
```diff
@@ -1801,3 +1802,40 @@ def extract(arch: any, source: list, lbl: str, forced_loop_type: any = None) ->
...
+
+
+def lookup_multidict(d, inst, find_class_func):
```
mkannwischer: Please add the Python documentation here.

slothy/targets/aarch64/aarch64_neon.py
```diff
@@ -2195,7 +2194,7 @@ class subs_imm(AArch64BasicArithmetic):
     modifiesFlags = True
 
 
-class sub_imm(AArch64BasicArithmetic):
+class sub_imm(AArch64Instruction):
```
mkannwischer: This looks like an unintentional change?

slothy/targets/aarch64/aarch64_neon.py
```diff
@@ -3644,7 +3643,7 @@ class Fmov(AArch64Instruction):
     pass
 
 
-class fmov_s_form(Fmov):
+class fmov_s_form(AArch64Instruction):
```
mkannwischer: This looks like an unintentional change?

mkannwischer (changes requested):
Thanks @willieyz - particularly for trying to reduce the code duplication!
Here are are couple of comments.

slothy/helper.py
willieyz: Hello @mkannwischer, thank you for taking the time to review!
I have corrected it to use the single class defined in `common.py`.

slothy/targets/aarch64/aarch64_big_experimental.py
willieyz: Fixed, now every call to `lookup_multidict()` takes a `instclass` as an input argument, for example:
```
def get_latency(src, out_idx, dst):
    instclass_src = find_class(src)
    latency = lookup_multidict(default_latencies, src, instclass_src)
    ....
```

slothy/targets/aarch64/aarch64_neon.py
willieyz: Thank you for point this out! 
This change is because of following reason:
In the neoverse_n1 model, both `sub_imm: 2` and `AArch64BasicArithmetic: 1` latencies were defined. However, according to the [Neoverse N1 SWOG (page 12)](https://developer.arm.com/documentation/109896/latest/), sub_imm should stick with Exec Latency 2.
Therefore, I changed the based class for `sub_imm` to separate these two cases, is this ok for our implementation?

slothy/targets/aarch64/aarch64_neon.py
willieyz: Similar to the previous case, `fmov_s_form` also conflicted with `Fmov`. The `fmov_s_form` is the vector-to-general register version and has different Exec Latency and Utilized Pipelines definitions from the other `Fmov` instruction in the `neoverse_n1` model, according to [Neoverse N1 SWOG (page 30)](https://developer.arm.com/documentation/109896/latest/)
Therefore, I updated the base class to separate these two cases.

slothy/helper.py
willieyz: Added.

slothy/targets/aarch64/aarch64_neon.py
mkannwischer: This sounds wrong though? 
```
sub x0, x1, x2
sub x0, x1, #1
```
is the basic variant and should have a latency of 1. 

```
sub x0, x1, x2, lsl #8
sub x0, x1, w2, SXTW
```
would be the extend/shift variants.

slothy/targets/aarch64/aarch64_neon.py
mkannwischer: I don't like that we change the _architectural model_ like this because of one _specific microarchitecture_.
Can we instead change the N1 microarchitectual model to not contain Fmov but only the sub-classes?
If you have a better class hierarchy proposal for all the fmov's, please let me know.

willieyz: I think your suggestion is the most reasonable one, I’ve updated the code accordingly. 
Thank you for your review and help!

slothy/targets/aarch64/aarch64_neon.py
willieyz: Yes, `sub_imm` should belong to the basic one, sorry about my mistake, I have corrected it, `sub_imm` now remains under `AArch64BasicArithmetic`.

PR #327, opened 2025-09-14
AArch64: Add eon instruction

FloydZ:
this pr adds the `eon` instruction to `aarch64` architecture and to the `a55`, `a72`, `neoverse N1` micro architectures. 
Note for `a72` I import the superset `AArch64Logical` instead of just `eon`, which includes all other logical instructions. And for N1 nothing had to be done, as it already imports `AArch64Logical`.

### DIFF
--- a/slothy/targets/aarch64/aarch64_neon.py
+++ b/slothy/targets/aarch64/aarch64_neon.py
@@ -2170,12 +2170,6 @@ class asr_wform(AArch64Instruction):
     outputs = ["Wd"]
 
 
-class eor_wform(AArch64Instruction):
-    pattern = "eor <Wd>, <Wa>, <Wb>"
-    inputs = ["Wa", "Wb"]
-    outputs = ["Wd"]
-
-
 class AArch64BasicArithmetic(AArch64Instruction):
     pass
 
@@ -2502,8 +2496,26 @@ class rev_w(AArch64Logical):
     outputs = ["Wd"]
 
 
+class eor_wform(AArch64Logical):
+    pattern = "eor <Wd>, <Wa>, <Wb>"
+    inputs = ["Wa", "Wb"]
+    outputs = ["Wd"]
+
+
+class eon_wform(AArch64Logical):
+    pattern = "eon <Wd>, <Wa>, <Wb>"
+    inputs = ["Wa", "Wb"]
+    outputs = ["Wd"]
+
+
 class eor(AArch64Logical):
     pattern = "eor <Xd>, <Xa>, <Xb>"
+    inputs = ["Xa", "Xb"]
+    outputs = ["Xd"]
+
+
+class eon(AArch64Logical):
+    pattern = "eon <Xd>, <Xa>, <Xb>"
     inputs = ["Xa", "Xb"]
     outputs = ["Xd"]
 
--- a/slothy/targets/aarch64/cortex_a55.py
+++ b/slothy/targets/aarch64/cortex_a55.py
@@ -119,6 +119,7 @@ from slothy.targets.aarch64.aarch64_neon import (
     bic,
     bic_reg,
     eor,
+    eon,
     ror,
     eor_shifted,
     bic_shifted,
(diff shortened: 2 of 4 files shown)


slothy/targets/aarch64/cortex_a55.py
```diff
@@ -491,7 +495,7 @@ def get_min_max_objective(slothy):
     (mov_d01, mov_b00): 1,
     (vzip1, vzip2): 1,
     (eor_wform): 1,
```
dop-amin: `eon_wform` missing from `inverse_throughput` on Cortex-A55.

slothy/targets/aarch64/cortex_a72_frontend.py
```diff
@@ -59,6 +59,7 @@
     find_class,
     all_subclass_leaves,
     AArch64ConditionalCompare,
+    AArch64Logical,
```
dop-amin, line 62: `eon_wform` does not inherit from `AArch64Logical`. Therefore, it's not considered in the Cortex-A72 model (same holds for Neoverse N1). I'd either add it here or leave out the w-form entirely.

tests/naive/aarch64/instructions.s
```diff
@@ -5,6 +5,8 @@ start:
 add x2, x1, #64
 add x2, x1, 64
 
+eon x2, x2, x19
```
dop-amin: As before, either adding a test for the w-form or omitting it in the models would be the way to go.

slothy/targets/aarch64/aarch64_neon.py
```diff
@@ -2177,6 +2177,12 @@ class eor_wform(AArch64Instruction):
     outputs = ["Wd"]
 
 
+class eon_wform(AArch64Instruction):
```
dop-amin: Not inheriting from `AArch64Logical` -- intentional?

dop-amin (commented):
Hi Floyd, 

thanks for submitting the PR! :-) I've left a few small comments but essentially, you're fulfilling the requirements we've asked for 🥃 

About the w-form: If you don't feel like applying the fixes for this, just go ahead and remove the changes regarding the w-form. Then this PR should be good to go!

Cheers from a former "Kryptographie" exercise group attendee

slothy/targets/aarch64/cortex_a72_frontend.py
FloydZ: this is on purpose. I did not want to add like ~10 instructions to the A72 and N1. For the A55 those instructions were already all imported, but separately.  I just did not want to blow up something by adding 10 unrelated instructions. But if wanted I can add them, thats not really the problem :)

slothy/targets/aarch64/aarch64_neon.py
FloydZ: yes, as the `eor_wform`

slothy/targets/aarch64/cortex_a55.py
FloydZ: oh good catch. I messed up the merge from main. thanks

slothy/targets/aarch64/cortex_a72_frontend.py
dop-amin: Hi Floyd, I'm not sure I'm following here. What would be the problem with either a) adding `eon_wform` to the A72/N1 models (just one instruction, exactly like for A55) or b) making it inherit from `AArch64Logical` (there is no need to add all child classes (concrete instructions) of a parent class if they all have the same microarchitectural characteristics)? I can see that `eor_wform` does not inherit from this class (as you mention below), which I believe is for historical reasons. 

Would be awesome if we could get this merged; either by including the wform in all three microarchitectural models and the test or by leaving it out entirely 🙂

tests/naive/aarch64/instructions.s
FloydZ: added w_form

slothy/targets/aarch64/cortex_a72_frontend.py
FloydZ: > What would be the problem with making it inherit from AArch64Logical

nothing :). moved the `eon_wform` and `eor_wform`. Hope thats alright.

dop-amin:
@FloydZ Thanks for the work! 👍 
I've cleaned up the history a bit and will merge your changes once CI passes one last time.

mkannwischer (approved):

PR #328, opened 2025-09-14
Add EON instruction.

J08nY:
Let's see what happens in CI... I have to find more info on the A72 and the N1. Getting the correct manuals is surprisingly tricky for someone who doesn't work in this space :)

Also, would it be possible to generate some of these configurations from machine-readable data on the instructions? I would hope it is out there somewhere. If not, at worst, PDF parsing may come in handy.


mkannwischer:
> Haha, and I am an hour late #327.

First passing the CI wins ;-)

J08nY:
> > Haha, and I am an hour late #327.
> 
> First passing the CI wins ;-)

Yeah, but I am not even sure of the correctness atm. :grin:

Issue #332, opened 2025-09-29
AArch64: Extend support for different loop variants


hanno-becker:
The AArch64 model currently only supports a very limited set of loop forms. To expand the range of programs that SLOTHY can be applied to without major modification, we should expand this.

For concrete examples, see the loop forms used in the assembly routines here: https://github.com/intel/isa-l/tree/master/erasure_code/aarch64

For example:
```asm
add     cnt, cnt, #128        # Increment position counter
...
cmp     cnt, threshold                  # Compare position vs length
ble     LABEL                       # Branch if less or equal (signed)
```

or

```asm
ldr/str   ..., [cnt], #128        # Increment position counter
...
cmp     cnt, threshold              # Compare position vs length
ble     LABEL                # Branch if less or equal (signed)
```

where `threshold` is fixed in either case.

@dop-amin How would you approach this?

dop-amin:
@hanno-becker I've addressed this in #333. Please take a look if that suits your needs.

PR #333, opened 2025-09-29
feat: BranchLoop for AArch64

dop-amin:
Addresses #332 
* Adds loop type that is less reliant on the structure of the loop itself
* Expects one compare instruction and one/multiple instructions modifying a counter used inside the compare.


hanno-becker (approved):
This looks good to me, thank you @dop-amin. I'll try it out on some complex workloads and let you know if there are any issues.

Issue #331, opened 2025-09-29
Add support for multiline comments


hanno-becker:
SLOTHY's assembly parse so far only looks out for `// ...`-style comments. We should support at least single-line `/* ... */` comments (should be easy with the current setup), though preferably also multi-line `/* ... */`.

PR #338, opened 2025-10-02
Add support for single-line and multi-line /* */ comments

mkannwischer:
This commit adds support for /* */ comments (in addition to the already supported // comments.
Within optimization regions these comments are removed just like // comments.

Outside of optimization regions, the comments are preserved. To ease parsing /* */ comments are first converted to // comments with additional markers to reconstruction in the end.

Resolves https://github.com/slothy-optimizer/slothy/issues/331


slothy/helper.py
```diff
@@ -232,7 +236,21 @@ def to_string(self, indentation=True, comments=True, tags=True):
...
+            # Convert comments with markers back to /* */ format
+            def format_comment(s):
+                if _NEWLINE_MARKER in s:
+                    # Restore as multi-line /* */ comment
+                    lines = s.split(_NEWLINE_MARKER)
+                    lines = [line.strip() for line in lines]
+                    return "/* " + "\n   ".join(lines) + " */"
+                elif s.startswith(_BLOCK_COMMENT_MARKER):
+                    # Restore as single-line /* */ comment
+                    return f"/* {s[len(_BLOCK_COMMENT_MARKER):]} */"
```
hanno-becker, line 248: It's unintuitive that different markers are used for block comments with and without line breaks. Why not always use BLOCK_COMMENT_MARKER at the start of a block comment, and within the block comment use _NEWLINE_MARKER for every newline.

hanno-becker (changes requested):
@mkannwischer Thanks a lot for working on this!

One nit regarding the internally. 

Functionally, we should strive to leave comments outside the optimization regions unchanged, including spacing. At the moment, though, the calls to `strip` break that. You can witness by adding for example
```c
/* this is a commen
 *   with weird
        * spacing
 */
```
at the top of your test file, and see it change in the optimized file.

hanno-becker (approved):
Actually, let's get this in and fix the spacing as a follow-up. This PR causes no regression for existing support for one-line comments, and is a step towards multiline support that is already of practical use, even if not perfect. We can iterate.

PR #344, opened 2025-10-13
Don't fail on missing llvm-mc

dop-amin:
* Disable selftest if llvm-mc cannot be found.
* Issue warning in log.

### DIFF
--- a/slothy/core/config.py
+++ b/slothy/core/config.py
@@ -28,7 +28,7 @@
 """
 SLOTHY configuration
 """
-
+import shutil
 
 from copy import deepcopy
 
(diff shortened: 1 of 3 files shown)


slothy/core/config.py
```diff
@@ -143,7 +143,14 @@ def selftest(self):
 
             This is so far implemented as a repeated randomized test -- nothing clever.
         """
-        return self._selftest
+        selftest_enabled = self._selftest
+        try:
+            subprocess.run(["llvm-mc", "--version"])
```
mkannwischer: Shall we switch this to `if shutil.which("llvm-mc") is None:`?

slothy/core/config.py
```diff
@@ -143,7 +143,14 @@ def selftest(self):
...
+        except FileNotFoundError:
+            self.logger.warning("llvm-mc not found")
```
mkannwischer: Warn the user here that Selftest will be disabled?

slothy/core/config.py
```diff
@@ -143,7 +143,12 @@ def selftest(self):
 
             This is so far implemented as a repeated randomized test -- nothing clever.
         """
-        return self._selftest
+        selftest_enabled = self._selftest
+        if shutil.which("llvm-mc") is None:
```
mkannwischer: The comments above suggest we also need `llvm-nm`, `llvm-readobj`? I'm not sure if those are always present if llvm-mc is present. It may be worthwhile to check for those too.

slothy/core/config.py
```diff
@@ -143,7 +143,12 @@ def selftest(self):
...
+            self.logger.warning("llvm-mc not found, disabling selftest.")
```
mkannwischer:
```suggestion
            self.logger.warning("LLVM (llvm-mc, llvm-nm, llvm-readobj) not found, disabling selftest. Consider installing LLVM.")
```

slothy/core/config.py
dop-amin: I don't know of a way to only install one and not the others but you're right, better to have that.

Issue #38, opened 2024-03-15
Problems with optimizing X25519


mkannwischer:
I was trying to put together an example showing how to use the splitting heuristic on the X25519. 
I'm not trying to have optimal code (for that we of course have https://github.com/slothy-optimizer/slothy/blob/main/paper/scripts/slothy_x25519.sh), but just trying to have something that terminates in a reasonable time on a laptop and demonstrates how the splitting heuristic works. 

I have the following code (to be placed in a Python script in the root dir):
```
import logging
import sys

from slothy import Slothy

import slothy.targets.aarch64.aarch64_neon as AArch64_Neon
import slothy.targets.aarch64.cortex_a55 as Target_CortexA55

logging.basicConfig(stream=sys.stdout, level=logging.INFO)

arch = AArch64_Neon
target = Target_CortexA55

slothy = Slothy(arch, target)

# example
slothy.load_source_from_file("examples/naive/aarch64/X25519-AArch64-simple.s")

slothy.config.inputs_are_outputs=True
slothy.config.outputs=["x0"]
slothy.config.variable_size=True
slothy.config.constraints.stalls_first_attempt=32
slothy.config.split_heuristic = True
slothy.config.split_heuristic_stepsize = 0.1
slothy.config.split_heuristic_factor = 10
slothy.config.split_heuristic_repeat = 2
slothy.optimize(start="mainloop", end="end_label")
slothy.write_source_to_file("test.s")
```
Running this very quickly errors with
```
INFO:slothy:Instructions in body: 958
INFO:slothy.mainloop.split.0_96:Perform internal binary search for minimal number of stalls...
INFO:slothy.mainloop.split.0_96:Attempt optimization with max 32 stalls...
INFO:slothy.mainloop.split.0_96:Objective: minimize number of stalls
INFO:slothy.mainloop.split.0_96:Invoking external constraint solver (OR-Tools CP-SAT v9.7.2996) ...
INFO:slothy.mainloop.split.0_96:[4.5895s]: Found 1 solutions so far... objective 12.0, bound 6.0 (minimize number of stalls)
INFO:slothy.mainloop.split.0_96:OPTIMAL, wall time: 6.632225 s
INFO:slothy.mainloop.split.0_96:Booleans in result: 8991
INFO:slothy.mainloop.split.0_96.selfcheck:OK!
INFO:slothy.mainloop.split.0_96:Minimum number of stalls: 12
[...]

PR #345, opened 2025-10-14
Add support for simplify ">>" and "<<" in <imm>

willieyz:
- During PR #341, we noticed that some shifted immediate values need to be handled.

- This PR adds support for simplifying shift operators (<< and >>) applied to immediate values.
   For example: `movs r0, #38 << 1` will be simplified to `movs r0, #76`.
- Apply on following architecture model:
  - `aarch_neon`
  - `arch_v7m`
  - `arch_v81m`

### DIFF
--- a/slothy/targets/aarch64/aarch64_neon.py
+++ b/slothy/targets/aarch64/aarch64_neon.py
@@ -1011,9 +1011,9 @@ class AArch64Instruction(Instruction):
         flag_pattern = "|".join(flaglist)
         dt_pattern = "(?:|2|4|8|16)(?:B|H|S|D|b|h|s|d)"
         imm_pattern = (
-            "(#(\\\\w|\\\\s|/| |-|\\*|\\+|\\(|\\)|=)+)"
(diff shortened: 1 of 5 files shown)


slothy/targets/aarch64/aarch64_neon.py
```diff
@@ -1220,6 +1220,10 @@ def build(c, src):
         depends_on_flags = getattr(c, "dependsOnFlags", False)
 
         if isinstance(src, str):
+            # Simplify the shifted imm, e.g., movs r0,#38<<1 simplified to movs r0,#76
+            src = re.sub(
+                r"#(\d+(?:<<|>>)\d+)", lambda m: f"#{simplify(m.group(1))}", src
+            )
```
mkannwischer: Why do we need to perform simplification here? We don't perform other simplifications such as `64 + 4`. 
Can't we just add `<<` and `>>` to the regular expression that matches the immediates and the existing calls to simplify will take care of it?

mkannwischer (changes requested):

PR #350, opened 2025-10-22
Armv7-M: Removing the wrong `args_in_out_different` for ldr/ldrb/ldrh

willieyz:
- We discovered this issue during #341, when we tried to optimize some assembly code that included `ldr` instructions. The `args_in_out_different` restriction made it harder for Slothy to find a solution in scenarios where the number of available registers was limited.

- This PR removes the restriction in the `make()` function of `ldr` and `ldr_with_imm`, namely the line:
  `obj.args_in_out_different = [(0, 0)] # Can't have Rd==Ra`.
  we believe this is wrong restriction for `ldr`, `ldr_imm`, `ldrh_with_imm` and `ldrb_with_imm`

### DIFF
--- a/slothy/targets/arm_v7m/arch_v7m.py
+++ b/slothy/targets/arm_v7m/arch_v7m.py
@@ -1854,7 +1854,6 @@ class ldr(Armv7mLoadInstruction):
         obj.increment = None
         obj.pre_index = 0
         obj.addr = obj.args_in[0]
-        obj.args_in_out_different = [(0, 0)]  # Can't have Rd==Ra
         return obj
 
     def write(self):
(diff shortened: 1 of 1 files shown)


slothy/targets/arm_v7m/arch_v7m.py
```diff
@@ -1875,7 +1874,6 @@ def make(cls, src):
         obj.increment = None
         obj.pre_index = obj.immediate
         obj.addr = obj.args_in[0]
-        obj.args_in_out_different = [(0, 0)]  # Can't have Rd==Ra
```
mkannwischer, line 1878: doesnät the same apply to `ldrb` and the other instrcutions below?

willieyz: Hello, @mkannwischer , 
thanks for your review, I had apply this change to other two instruction below(`ldrh_with_imm` and `ldrb_with_imm`), thank your for your review.

mkannwischer (approved):
LGTM. Thanks @willieyz.

PR #355, opened 2025-10-30
Add scalar movz to aarch64

### LINKED ISSUE #353: Add support for AArch64 scalar move instructions
From what I can see, the below instructions are not supported, which are helpful in optimizing scalar crypto code:

- `movz <Xd>, <imm>`
- `movz <Xd>, <imm>, lsl <shift>`

dop-amin:
Resolves #353 
Based on #354


slothy/targets/aarch64/aarch64_neon.py
```diff
@@ -2005,6 +2005,24 @@ def write(self):
...
 
 
+class x_ldr_regidx(Ldr_X):
+    pattern = "ldr <Xa>, [<Xb>, <Xc>]"
+    inputs = ["Xb", "Xc"]
+    outputs = ["Xa"]
+
+    @classmethod
+    def make(cls, src):
+        obj = AArch64Instruction.build(cls, src)
```
hanno-becker, line 2015: Remind me, what determines whether an instruction participates in address offset fixup?

dop-amin: Judging from the documentation [here](https://github.com/slothy-optimizer/slothy/blob/9d9d9d61f030b98272d2c539f1c5b1f37acf9439/slothy/targets/aarch64/aarch64_neon.py#L2053), asserting on the `pre_index` will make the optimization fail if it was fixed-up. 

More generally, if the instruction is not a load/store it will not be subject to fixup (`is_load_store_instruction` is being checked). 
As `x_ldr_regidx` has `Ldr_X` as its base class, it will be a load/store instruction. Not marking it as such may have implications in other parts of the code. 

If this instruction is used in combination with instructions that DO increment the same pointer, a non-zero fixup may be applied (see [here](https://github.com/slothy-optimizer/slothy/blob/b8601592d0cf1a848e82be5420b178f7cce939a5/slothy/core/core.py#L1299)). 

I have not seen any code before that mixes register offset with immediate increments so far but it could exist of course.

PR #356, opened 2025-10-31
Fix writing to undefined log location

dop-amin:
* Add log_dir config option
* Add test to check that a test that's failing because it's impossible to optimize within the constraints (-> No solution found) actually fails for this reason.

### DIFF
--- a/slothy/core/config.py
+++ b/slothy/core/config.py
@@ -830,6 +830,16 @@ class Config(NestedPrint, LockAttributes):
         """
         return self._log_dir
 
+    @property
+    def log_model(self):
+        """Whether and where to export the constraint solver model.
+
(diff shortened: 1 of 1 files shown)


slothy/core/config.py
```diff
@@ -1397,6 +1405,8 @@ def __init__(self, Arch, Target, logger):
         self.solver_random_seed = 42
 
         # TODO: Document log_dir and log_model
```
mkannwischer: While at it, could you address this TODO?

mkannwischer: I fixed it myself.

mkannwischer (approved):
LGTM. Thanks @dop-amin!

PR #372, opened 2025-12-11
[core] Allow if/else asm directives in optimize_loop

dop-amin:
* Resolving of if/else directives was not previously done in optimize_loop. Added.
* Added test case.


mkannwischer (approved):
LGTM. Good catch.

PR #375, opened 2025-12-16
aarch64: Proper fcsel support

dop-amin:
* Add fcsel as an instruction, not just a wrapper
* Add to uArch models
* Switch X25519 from wrapper to fcsel instruction

I have not yet run the full X25519 optimization and re-run the tests. Do we require this @mkannwischer?

### DIFF
--- a/examples/naive/aarch64/x25519/X25519-AArch64-simple.s
+++ b/examples/naive/aarch64/x25519/X25519-AArch64-simple.s
@@ -29,10 +29,6 @@
 #include <hal_env.h>
 #include "instruction_wrappers.i"
 
-.macro fcsel_dform out, in0, in1, cond // @slothy:no-unfold
-  fcsel dform_\out, dform_\in0, dform_\in1, \cond
-.endm
-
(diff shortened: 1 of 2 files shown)


slothy/targets/aarch64/aarch64_neon.py
```diff
@@ -2923,6 +2923,21 @@ class cset(AArch64ConditionalSelect):
...
 
+class fcsel(AArch64ConditionalSelect):
+    """FCSEL - Floating-point Conditional Select (D-form scalar)
+
+    Syntax: FCSEL <Dd>, <Dn>, <Dm>, <cond>
+
+    Selects between two double-precision floating-point values based on
+    condition flags. Executes on SIMD/FP units.
+    """
+
```
mkannwischer: We don't usually add docstrings here.  Why do we start it now?

dop-amin: Too many projects with different requirements 🙂 I've removed it.

PR #385, opened 2026-01-02
AArch64: Add lane-indexed st2 variants to Arch model

mkannwischer:


slothy/targets/aarch64/aarch64_neon.py
```diff
@@ -754,7 +754,7 @@ def is_q_form_vector_instruction(self):
             return self._is_instance_of([Str_Q, Ldr_Q])
 
         # Operations on specific lanes are not counted as Q-form instructions
-        if self._is_instance_of([Q_Ld2_Lane_Post_Inc]):
+        if self._is_instance_of([Q_Ld2_Lane_Post_Inc, st2_lane, st2_lane_post_inc]):
```
dop-amin, line 757: Just an idea for the future: We could also make `is_lane_mode` a property of the instruction so it's not as easy to overlook adding it here. 
The change itself here is fine of course.

slothy/targets/aarch64/aarch64_neon.py
```diff
@@ -1563,6 +1563,7 @@ def make(cls, src):
         obj.args_in_out_combinations = [
             ([0, 1], [[f"v{i}", f"v{i+1}"] for i in range(0, 30)])
         ]
+        obj.addr = obj.args_in_out[2]
```
dop-amin, line 1566: Nit:
This change is unrelated to `st2`, right? 
If so, could you update the commit message to include this/put in separate commit?

slothy/targets/aarch64/neoverse_n1_experimental.py
```diff
@@ -342,6 +345,7 @@ def get_min_max_objective(slothy):
     vuaddlv_sform: 5,  # 8B/8H
     q_ldr1_stack: 7,
     Q_Ld2_Lane_Post_Inc: 7,
+    St2: 2,
```
dop-amin: What's the source for this latency?
N1 SWOG page 55 says 4 cycles for lane-mode, doesn't it?

slothy/targets/aarch64/neoverse_n1_experimental.py
```diff
@@ -238,6 +239,7 @@ def get_min_max_objective(slothy):
     vuaddlv_sform: [[ExecutionUnit.VEC0, ExecutionUnit.VEC1]],
     q_ldr1_stack: ExecutionUnit.V(),
     Q_Ld2_Lane_Post_Inc: ExecutionUnit.V(),
+    St2: ExecutionUnit.V(),
```
dop-amin: Same as with `st3` and `st4`, the L pipe should also be occupied. Might be worth moving this class further to the top to relate the comment on this matter to it as well?

slothy/targets/aarch64/aarch64_neon.py
mkannwischer: Moved to a second commit.

slothy/targets/aarch64/neoverse_n1_experimental.py
mkannwischer: Good catch. Fixed.

slothy/targets/aarch64/neoverse_n1_experimental.py
mkannwischer: Moved it further to the top.

slothy/targets/aarch64/aarch64_neon.py
mkannwischer: Sounds useful. Do you have time to add it in separate PR?

Issue #230, opened 2025-05-30
Make naming of stack locations configurable


hanno-becker:
Currently, when SLOTHY is used in stack-spilling mode, it'll use STACK_LOC_XXX for stack locations. This should be configurable.

PR #386, opened 2026-01-05
Make naming of stack locations configurable

willieyz:
- Resolves: #230
- The commit allows users to specify a custom prefix for stack location names during spill/restore operations when a specific naming convention is required.
- Currently, this change has been tested via the mlkem-native PR:
  https://github.com/pq-code-package/mlkem-native/pull/1396, 
  and the optimization ran successfully using the MLK_STACK_LOC_ prefix.

### DIFF
--- a/slothy/core/config.py
+++ b/slothy/core/config.py
@@ -1101,12 +1101,28 @@ class Config(NestedPrint, LockAttributes):
             """Allow Slothy to introduce stack spills
 
             When this option is enabled, Slothy will consider the introduction
-            of stack spills to reduce register pressure.
+            of stack spills to reduce register pressure. The naming of stack
+            locations can be customized using spill_stack_loc_prefix.
 
             This option should only be disabled if it is known that the input
             assembly suffers from high register pressure. For example, this can
             be the case for symbolic input assembly."""
             return self._allow_spills
+
+        @property
+        def spill_stack_loc_prefix(self):
+            """
+            Prefix for stack location names in generated assembly when spills
+            are enabled.
+
+            When allow_spills is True, this prefix is used to generate unique stack
+            location names in the format "{prefix}_{index}" (e.g., "STACK_LOC_0",
+            "STACK_LOC_1", etc.). This allows customization of naming conventions
+            to match project requirements.
+
+            Default: "STACK_LOC"
+            """
+            return self._spill_stack_loc_prefix
 
         @property
         def spill_type(self):
@@ -1190,6 +1206,7 @@ class Config(NestedPrint, LockAttributes):
             self._allow_reordering = True
             self._allow_renaming = True
             self._allow_spills = False
+            self._spill_stack_loc_prefix = "STACK_LOC"
             self._spill_type = None
 
             self.lock()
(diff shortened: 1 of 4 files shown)


slothy/cli/__init__.py
```diff
@@ -96,6 +96,11 @@ def main():
     )
     parser.add_argument("--fusion", default=False, action="store_true")
     parser.add_argument("--fusion-only", default=False, action="store_true")
+    parser.add_argument(
+        "--stack-loc-prefix",
+        default="STACK_LOC",
+        help="Prefix for stack location names in generated assembly (default: STACK_LOC)",
+    )
```
mkannwischer: Why does this need a separate command line argument? 
Can't we just pass it using the usual `-c stack_loc_prefix=`?

slothy/core/config.py
```diff
@@ -606,6 +606,11 @@ def has_objective(self):
     def absorb_spills(self):
         return self._absorb_spills
 
+    @property
+    def stack_loc_prefix(self):
```
mkannwischer: Rename this to `spill_stack_log_prefix`, so it is clearer this is related to spills.
can you move this option just below allow_spills, so that the documentation is close to each other.

slothy/core/config.py
```diff
@@ -606,6 +606,11 @@ def has_objective(self):
...
+        """Prefix for stack location names in generated assembly"""
```
mkannwischer: Extend this documentation to explain what it actually does and what the default behaviour is.
The `allow_spills` documentation should also be extended to refer to this option.

slothy/core/core.py
```diff
@@ -635,7 +635,7 @@ def _get_code(self, visualize_reordering):
         def gen_restore(reg, loc, vis):
             args = self.config.constraints.spill_type
             yield SourceLine(
-                self.config.arch.Spill.restore(reg, loc, **args)
+                self.config.arch.Spill.restore(reg, loc, config=self.config, **args)
```
mkannwischer: Do we need to pass the entire config? just passing the naming prefix feels more natrual to me.

slothy/core/core.py
```diff
@@ -644,9 +644,11 @@ def gen_restore(reg, loc, vis):
 
         def gen_spill(reg, loc, vis):
             args = self.config.constraints.spill_type
-            yield SourceLine(self.config.arch.Spill.spill(reg, loc, **args)).set_length(
-                self.fixlen
-            ).set_comment(vis).add_tag("is_spill", True).add_tag(
+            yield SourceLine(
+                self.config.arch.Spill.spill(reg, loc, config=self.config, **args)
```
mkannwischer: same here.

mkannwischer (changes requested):
Thanks @willieyz. Overall this functions well.
A couple of comments below.

slothy/cli/__init__.py
willieyz: Fixed

slothy/core/config.py
willieyz: Fixed

slothy/core/config.py
willieyz: Fixed

slothy/core/core.py
willieyz: Fixed

PR #397, opened 2026-01-15
Update ortools to 9.15.6755 for all Python versions

mkannwischer:
Add Python 3.14 to test matrix and classifiers. Remove version-specific ortools dependencies and workarounds for older versions.


mkannwischer:
The README still needs updating.

PR #407, opened 2026-01-30
Correct vmull (polynomial) model on Cortex-M85 r1

willieyz:
This PR corrects the execution model of vmull (polynomial) according
to the Cortex-M85 SWOG (MVE integer vector instructions).

The correct model is:

  - ExecutionUnit: Group A
  - Latency: 1
  - Inverse throughput: 2

The polynomial form must be modeled separately from vmull (integer)
according to the souceLine' s datatype
This PR adds three helper functions:

  - is_dt_form_of()

This helper function are referenced from `aarch64_neon.py` and distinguish
between vmull (integer) and vmull (polynomial) based on the operand
data type.

### DIFF
--- a/slothy/targets/arm_v81m/cortex_m85r1.py
+++ b/slothy/targets/arm_v81m/cortex_m85r1.py
@@ -186,6 +186,7 @@ from slothy.targets.arm_v81m.arch_v81m import (
     ldrb_no_imm,
     ldrb_with_writeback,
     ldrb_with_post,
+    is_dt_form_of,
 )
 
 issue_rate = 1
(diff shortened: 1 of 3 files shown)


slothy/targets/arm_v81m/cortex_m85r1.py
```diff
@@ -205,6 +206,7 @@ class ExecutionUnit(Enum):
     VEC_IVMINMAX = (3,)
     VEC_SHFT = (3,)
     VEC_VMOVLN = (3,)
+    VEC_VMULLP = (3,)
```
mkannwischer: No need to add this. Just use VEC_INT

slothy/targets/arm_v81m/cortex_m85r1.py
```diff
@@ -687,6 +689,8 @@ def get_min_max_objective(slothy):
         rsb_imm,
         and_imm,
         sbfx,
+        is_polyform_form_of(vmullb),
```
mkannwischer: Wont this lead to multiple matches because the regular vmullb is also still in the dict?

slothy/targets/arm_v81m/cortex_m85r1.py
```diff
@@ -186,6 +186,7 @@
     ldrb_no_imm,
     ldrb_with_writeback,
     ldrb_with_post,
+    is_polyform_form_of,
```
mkannwischer: Probably, this should be generalized to something that can model both poly and not poly

slothy/targets/arm_v81m/cortex_m85r1.py
```diff
@@ -894,6 +898,11 @@ def get_latency(src, out_idx, dst):
 
 def get_units(src):
     instclass_src = find_class(src)
+    # Special handling for polynomial vmull
+    if isinstance(src, (vmullb, vmullt)) and is_polyform_form_of(type(src))(src):
```
mkannwischer: Why is this special handling needed. We should be able to use the is_form_of function also in the execution unit dict.

Issue #412, opened 2026-02-06
arch_v81m: add `orr`, `ldrb`, `add` instruction support


willieyz:
Required to add following instruction support

```
slothy.targets.arm_v81m.arch_v81m.Instruction.ParsingException: Couldn't parse orr r5, r3, r5, lsl#4
slothy.targets.arm_v81m.arch_v81m.Instruction.ParsingException: Couldn't parse ldrb r5, [r12, r5]
slothy.targets.arm_v81m.arch_v81m.Instruction.ParsingException: Couldn't parse add r5, r13, r5, lsl#5
```

PR #414, opened 2026-02-09
Armv8.1-M: Add orr, ldrb, add, ubfx instruction support

willieyz:
- Resolves: #412 

This PR add following 4 instruction:

- ubfx
  - m55 SWOG (page: Arithmetic and Logical instructions)
    - latency: 1
    - Inverse throughput: 1 (same as sbfx)
    - ExecutionUnit: SCALAR (same as sbfx)

  - m85 SWOG (page: Arithmetic and Logical instructions)
    - latency: 1
    - Inverse throughput: 1 (same as sbfx)
    - ExecutionUnit: SCALAR (same as sbfx)

- add (register, T3, `lsl` usage)
  - m55 SWOG (page: Arithmetic and Logical instructions)
    - latency: 2
      (We don't considerate the condition: shift amount as zero)
    - Inverse throughput: 1
    - ExecutionUnit: SCALAR (same as add)

  - m85 SWOG (page: Arithmetic and Logical instructions)

    - latency: 1(2)
      (according to SWOG's NOTE 1, the latency from the shifter source
       operand is 2)
    - Inverse throughput: 1
    - ExecutionUnit: SCALAR (same as add)

- ldrb (register, T1)
  - m55 SWOG (page: Load instructions)
    - latency: 2
    - Inverse throughput: 1
    - ExecutionUnit: LOAD

  - m85 SWOG (page: Load instructions)
    - latency: 3
    - Inverse throughput: 1
    - ExecutionUnit: LOAD

- orr_lsl (orr, register, T2)
- m55 SWOG (page: 20/48)
  - latency: 1(2)
    (According SWOG' note 2: If the shift amount is non-zero, the
     latency is 2 cycles and the throughput is 1. Since we never use a
     shift with a value of 0, we model this as 2 cycles, and add a note
     indicating that the latency would be 1 cycle if the shift
     amount were 0.)
  - Inverse throughput: 1
  - ExecutionUnit: SCALAR

- m85 SWOG (page: 16/50)
  - latency: 1
    (according to SWOG's NOTE 1, the latency from the shifter source
     operand is 2)
  - Inverse throughput: 1
  - ExecutionUnit: SCALAR

- orr (register, T1)
- m55 SWOG (page: 23/48)
  - latency: 1
  - Inverse throughput: 1
  - ExecutionUnit: SCALAR

- m85 SWOG (page: 21/50)
  - latency: 1
  - Inverse throughput: 1
  - ExecutionUnit: SCALAR

- orr_immediate (immediate)
- m55 SWOG (page: 20/48)
  - latency: 1
  - Inverse throughput: 1
  - ExecutionUnit: SCALAR

- m85 SWOG (page: 21/50)
  - latency: 1
  - Inverse throughput: 1
  - ExecutionUnit: SCALAR


slothy/targets/arm_v81m/arch_v81m.py
```diff
@@ -989,6 +989,12 @@ class add(MVEInstruction):
     outputs = ["Rd"]
 
 
+class add_lsl(MVEInstruction):
+    pattern = "add <Rd>, <Rn>, <Rm>, lsl <imm>"
```
mkannwischer, line 993: We should implement that via `<barrel>` as in the Armv7-M. I have opened https://github.com/slothy-optimizer/slothy/issues/416 to do this as a follow-up.

slothy/targets/arm_v81m/cortex_m55r1.py
```diff
@@ -757,6 +761,10 @@ def get_latency(src, out_idx, dst):
     # Check for latency exceptions
     #
 
+    # inst_src -> dst, if the shift amount is non-zero, then the latency is 2.
+    if instclass_src in [add_lsl] and (int(src.immediate) != 0):
+        return 2
+
```
mkannwischer: I think it is not worth to model this separately. We virtually never use a shift with 0 distance. Let's just use 2 and add a note that it would be 1 if it's zero.

slothy/targets/arm_v81m/cortex_m55r1.py
```diff
@@ -601,6 +603,7 @@ def get_min_max_objective(slothy):
         vfma,
         vmov_vector,
     ): 2,
+    orr_lsl: 1,
```
mkannwischer: Can you not add this as an seprate section, but keep it with the other basic ALU instructions.
While you are at it you can also add the basic `orr` instruction.

slothy/targets/arm_v81m/cortex_m55r1.py
```diff
@@ -752,6 +755,7 @@ def get_min_max_objective(slothy):
         vfma,
     ): 2,
     vmlaldava: 3,
+    orr_lsl: 1,
```
mkannwischer: same here.

Issue #415, opened 2026-02-09
Armv8.1-M: Add ldr, mov, mul and eor instruction support


willieyz:
The following instruction are required to be support by slothy

~slothy.targets.arm_v81m.arch_v81m.Instruction.ParsingException: Couldn't parse ldr.w r5, [r10], #4~
~slothy.targets.arm_v81m.arch_v81m.Instruction.ParsingException: Couldn't parse mov.w r9, #0x01010101~
```
slothy.targets.arm_v81m.arch_v81m.Instruction.ParsingException: Couldn't parse mul  r6, r5, r8
slothy.targets.arm_v81m.arch_v81m.Instruction.ParsingException: Couldn't parse and r8, r9, r5, lsr #7
slothy.targets.arm_v81m.arch_v81m.Instruction.ParsingException: Couldn't parse eor r5, r5, r8, lsl #7
```

PR #417, opened 2026-02-13
Armv8.1-M: Add ldr, mov, mul and eor instruction support

willieyz:
- Resolves: #415 

This PR add following instruction:

- `mul`
    - m55 SWOG (page: 27/48)
      - latency: 2
      - Inverse throughput: 1
      - ExecutionUnit: SCALAR

    - m85 SWOG (page: 23/50)
      - latency: 2
      - Inverse throughput: 1
      - ExecutionUnit: SCALAR

- `and_lsr`(`and`, register, T1 and  T2 variant, usage for lsr)
  - m55 SWOG (page: Arithmetic and Logical instructions)
     - latency: 1(2)
        (According SWOG' note 2:If the shift amount is non-zero, the
        latency is 2 cycles and the throughput is 1. Since we never use a
        shift with a value of 0. we model this as 2 cycles, and add a note
        indicating that the latency would be 1 cycle if the shift amount
        were 0.)
    - Inverse throughput: 1
    - ExecutionUnit: SCALAR (same as add_lsl)

  - m85 SWOG (page: Arithmetic and Logical instructions)
    - latency: 1(2)
      (according to SWOG's NOTE 1, the latency from the shifter source
       operand is 2)
    - Inverse throughput: 1
    - ExecutionUnit: SCALAR (same as add_lsl)

  - This PR also add the `and` (register T1)
  
  - m55 SWOG (page: Arithmetic and Logical instructions)
    - latency: 1
    - Inverse throughput: 1
    - ExecutionUnit: SCALAR (same as orr)
  
  - m85 SWOG (page: Arithmetic and Logical instructions)
    - latency: 1
    - Inverse throughput: 1
    - ExecutionUnit: SCALAR (same as orr)

- `eor_lsl` (`eor`, register, T1 and T2 variant, usage for `lsl`)
  - m55 SWOG (page: Arithmetic and Logical instructions)
     - latency: 1(2)
        (According SWOG' note 2:If the shift amount is non-zero, the
        latency is 2 cycles and the throughput is 1. Since we never use a
        shift with a value of 0. we model this as 2 cycles, and add a note
        indicating that the latency would be 1 cycle if the shift amount
        were 0.)
    - Inverse throughput: 1
    - ExecutionUnit: SCALAR (same as add_lsl)

  - m85 SWOG (page: Arithmetic and Logical instructions)
    - latency: 1(2)
      (according to SWOG's NOTE 1, the latency from the shifter source
       operand is 2)
    - Inverse throughput: 1
    - ExecutionUnit: SCALAR (same as add_lsl)

  - This commit also add the `eor` (register T1)
  
  - m55 SWOG (page: Arithmetic and Logical instructions)
    - latency: 1
    - Inverse throughput: 1
    - ExecutionUnit: SCALAR (same as orr)
  
  - m85 SWOG (page: Arithmetic and Logical instructions)
    - latency: 1
    - Inverse throughput: 1
    - ExecutionUnit: SCALAR (same as orr)


slothy/targets/arm_v81m/arch_v81m.py
```diff
@@ -995,6 +995,24 @@ class add_lsl(MVEInstruction):
     outputs = ["Rd"]
 
 
+class and_twoarg(MVEInstruction):
```
mkannwischer: Maybe we stick to the same name as in the Armv7-M model: `log_and`?

willieyz: Hello, @mkannwischer , 
Thank you for your review, I had change it to `log_and`
Thanks!

Issue #416, opened 2026-02-09
Armv8.1-M: Port `<barrel>`  from Armv7-M for barrel-shifted instructions


mkannwischer:
In the Armv7-M model, we have introduced the `<barrel>` shortcut for defining barrel-shifted instruction, e.g., 
```py
class add_shifted(Armv7mShiftedArithmetic):
    pattern = "add<width> <Rd>, <Ra>, <Rb>, <barrel><imm>"
    inputs = ["Ra", "Rb"]
    outputs = ["Rd"]
```

In the Armv8.1-M model we are still modelling this manually, e.g., via
```py
class add_lsl(MVEInstruction):
    pattern = "add <Rd>, <Rn>, <Rm>, lsl <imm>"
    inputs = ["Rn", "Rm"]
    outputs = ["Rd"]
```

We should add `<barrel>` support to the Armv8.1-M model and use it for existing instructions.

PR #418, opened 2026-02-24
Armv8.1-M: Port <barrel> from Armv7-M for barrel-shifted instructions

willieyz:
- Resolves: #416

This PR ports the <barrel> shortcut from armv7m to the armv81m model, refactors instruction patterns to replace explicit shift operands (lsl/ror/lsr/asr) with <barrel>, and renames the affected pattern classes following the naming convention used in armv7m.


mkannwischer (approved):
Thanks @willieyz!

PR #427, opened 2026-03-23
Add X25519 example in python and CI

dop-amin:
This PR is supposed to improve upon the CI coverage and completeness fo examples of the assembly present in the repo.

Currently done:
- x25519 on aarch64


mkannwischer (approved):
LGTM. Did you benchmark this on an A53/A55 to check if this achieves the same performance as in the paper?

dop-amin:
The benchmarks don't fully match the original performance. 
I've done two optimization runs on one of our big servers (256 cores AMD Epyc) and got these numbers from the resulting code: 
`Run 1: 141494`
`Run 2: 142008`
The original was 139752.

As there appears to be some variance in the optimization (+500 cycles from Run 1 -> Run 2), the question is whether the difference to the original is due to that or due to *some* change in the model/SLOTHY. I will try running the optimization a few more times, I recall we took the best result over multiple runs back then as well.

PR #434, opened 2026-04-03
Fix typo: vars -> bvars in _add_path_constraint_from

dop-amin:
`vars` is a Python built-in function. Should be the variable also used in the surounding code, `bvars`.


mkannwischer (approved):
LGTM.

PR #437, opened 2026-04-17
Prefer caller-save registers and emit callee-save clobber comments

dop-amin:
* Add `callee_saved_registers()` to Arm v8.1-M/Helium and RISC-V ISA models
* Add `config.constraints.prefer_caller_save_registers`: biases allocation toward input registers and away from callee-saved ones via CP-SAT hints and a secondary minimisation objective (never sacrifices stall counts)
* Add `config.emit_clobbered_callee_saves_comment` (default True): prepends a comment listing clobbered callee-saved registers to the output
* Accumulate clobber set across preamble, kernel, and postamble in `Heuristics.periodic`; store on `Result`
* Add four AArch64 naive tests covering the new features including a forced callee-save allocation case and a SW-pipelining path

### DIFF
--- a/slothy/core/config.py
+++ b/slothy/core/config.py
@@ -530,6 +530,23 @@ class Config(NestedPrint, LockAttributes):
         return self._with_llvm_mca_after
 
     @property
+    def emit_clobbered_callee_saves_comment(self):
+        """If True, prepend a comment to the output listing callee-saved registers
+        (per the ISA calling convention) that are clobbered by the optimized code.
+
+        This is useful for understanding the ABI cost of a snippet and for
+        constructing correct save/restore sequences around it.
+
+        Requires the ISA model to define callee_saved_registers(). When the ISA
+        model does not provide this method the option has no effect.
+        """
+        return self._emit_clobbered_callee_saves_comment
+
+    @emit_clobbered_callee_saves_comment.setter
+    def emit_clobbered_callee_saves_comment(self, val):
+        self._emit_clobbered_callee_saves_comment = val
+
+    @property
     def compiler_binary(self):
         """The compiler binary to be used.
 
@@ -619,6 +636,7 @@ class Config(NestedPrint, LockAttributes):
                 self.constraints.move_stalls_to_bottom is True,
                 self.constraints.minimize_register_usage is not None,
                 self.constraints.minimize_use_of_extra_registers is not None,
+                self.constraints.prefer_caller_save_registers is True,
                 self.target.has_min_max_objective(self),
             ]
         )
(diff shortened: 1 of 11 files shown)


slothy/core/config.py
```diff
@@ -1220,6 +1238,35 @@ def minimize_spills(self):
...
+            objective, optimal scheduling is never sacrificed: if callee-saved
+            registers are the only way to achieve minimum stalls, they will still
+            be used.
+
+            This option subsumes hints.rename_hint_orig_rename: when both are
+            enabled the orig-rename hint is already covered, so enabling
+            rename_hint_orig_rename on top is harmless but redundant.
+
+            Requires the ISA model to define callee_saved_registers(). When the
+            ISA model does not provide this method the option has no effect.
```
mkannwischer, line 1262: Can we make it error instead? Sounds easy to misuse otherwise.

slothy/core/core.py
```diff
@@ -2476,6 +2489,17 @@ def get_code_line(line_no):
...
                 self.logger.result.debug("> " + s.to_string())
 
+    def _compute_clobbered_callee_saved(self):
+        if not hasattr(self.arch.RegisterType, "callee_saved_registers"):
+            return set()
+        callee_saved = set(self.arch.RegisterType.callee_saved_registers())
+        used = set()
+        for t in self._get_nodes():
+            used.update(t.inst.args_out)
+            used.update(t.inst.args_in)
```
mkannwischer, line 2499: Why are `args_in` considered clobbered?

slothy/core/core.py
```diff
@@ -3977,6 +4040,15 @@ def _add_objective(self, force_objective=False):
...
 
+        # Secondary objective: prefer caller-save registers.
+        # This runs only in the force_objective (step-2 re-optimization) path,
+        # where stalls are already fixed via a hard constraint.
+        if force_objective and self._model._callee_saved_used and len(maxlist) == 0:
+            callee_vars = list(self._model._callee_saved_used.values())
+            if len(minlist) == 0:
+                minlist = callee_vars
+                name = "minimize callee-saved register usage"
+
```
mkannwischer, line 4051: What happens here if you combine it with another secondary objective, e.g., maximize_register_lifetimes?

slothy/core/slothy.py
```diff
@@ -116,6 +116,17 @@ def __init__(self, arch, target, logger=None):
         self.last_result = None
         self.success = None
 
+    def _prepend_clobber_comment(self, code, clobbered, indentation):
+        """Prepend a callee-saved clobber comment to code if the feature is on."""
+        if not (self.config.emit_clobbered_callee_saves_comment and clobbered):
+            return code
+        comment_text = "Clobbered callee-saved registers: " + ", ".join(
+            sorted(clobbered)
```
mkannwischer, line 124: sorting lexicographically is probably not what you want to do here. Can we instead use a more natural order?

mkannwischer (changes requested):
Thanks @dop-amin. Looks like a useful feature! It's not very clear how it interacts with other secondary objectives to me. Have you considered that?

PR #441, opened 2026-05-06
AArch64: add CRC32/CRC32C instruction family to (A55, A72, N1) models

ckocaogullar:
Add all eight scalar CRC32/CRC32C variants (B/H/W/X) to the AArch64 architecture model and Neoverse N1 microarchitecture model.

Scheduling data (Neoverse N1 Software Optimization Guide, section 3.21):
- Execution unit: M
- Inverse throughput: 1
- Latency: 2 (1 with CRC->CRC chain forwarding, see Note 1 in SWOG section 3.21)

Also adds crc to llvm_mc_attr so llvm-mc can assemble CRC32 instructions during selftest.

### DIFF
--- a/slothy/targets/aarch64/cortex_a55.py
+++ b/slothy/targets/aarch64/cortex_a55.py
@@ -181,6 +181,14 @@ from slothy.targets.aarch64.aarch64_neon import (
     fmov_s_form,  # from double/single to gen reg
     cmp,
     vdup_w,
+    crc32b,
+    crc32h,
+    crc32w,
+    crc32x,
+    crc32cb,
+    crc32ch,
+    crc32cw,
+    crc32cx,
 )
 
 issue_rate = 2
@@ -448,6 +456,16 @@ execution_units = {
     # is not modeled
     AESInstruction: [[ExecutionUnit.VEC0, ExecutionUnit.VEC1]],
     csel: ExecutionUnit.SCALAR(),
+    (
+        crc32b,
+        crc32h,
+        crc32w,
+        crc32x,
+        crc32cb,
+        crc32ch,
+        crc32cw,
+        crc32cx,
+    ): ExecutionUnit.SCALAR(),
 }
 
 inverse_throughput = {
@@ -544,6 +562,7 @@ inverse_throughput = {
     fmov_s_form: 1,  # from double/single to gen reg
     vdup_w: 1,
     (sxtb, uxtb): 1,
+    (crc32b, crc32h, crc32w, crc32x, crc32cb, crc32ch, crc32cw, crc32cx): 1,
 }
 
 default_latencies = {
@@ -652,6 +671,8 @@ default_latencies = {
     AESInstruction: 2,
     fmov_s_form: 1,  # from double/single to gen reg
     (sxtb, uxtb): 1,
+    (crc32b, crc32h, crc32x, crc32cb, crc32ch, crc32cx): 2,
+    (crc32w, crc32cw): 1,
 }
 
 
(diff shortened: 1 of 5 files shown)


examples/naive/aarch64/crc32/crc32_all.s
```diff
@@ -0,0 +1,49 @@
...
+    crc32b  w5, w5, w7
+    crc32h  w5, w5, w7
+    crc32w  w5, w5, w7
+    crc32x  w5, w5, x8
+
+    /// Castagnoli CRC32C variants (polynomial 0x1EDC6F41).
+    crc32cb w6, w6, w7
+    crc32ch w6, w6, w7
+    crc32cw w6, w6, w7
+    crc32cx w6, w6, x8
```
mkannwischer: Is there a reason why those can't be added to our default instructions.s test?
Then we wouldn't need a new test for it.

ckocaogullar: My understanding is that the `instructions.s` test runs against A55, A72, and N1, and as these changes only add CRC32 modeling for N1, adding it there would cause `UnknownInstruction` failures on the other two targets. Let me know if I'm missing something and I'll make the change.

mkannwischer: It shouldn't be an `UnknownInstruction` error, but yes one would run into an error if you don't add the new instructions to the A55 and A72 model as well.
Could you kindly do that too? Below are the corresponding performance characteristics from the SWOGs. Let me know if you need any help.

A72: 
image

latency 2 - with a forwarding path between two crc instructions. 
throughput 1. 
All instructions using the multi-cycle pipeline have throughput 1, so I think the current modelling of `MINT0` and `MINT1` is wrong - it should be changed to just one `MINT` pipe. 

A55: 

image

No forwarding path on A55 according to the SWOG. 
Latency depends on the instruction used, so you'll have to list each individual instruction.
It doesn't say anything about the execution unit, but we can model it to run on either of the two ALU units.
Each of them will have an inverse throughput of 1.

ckocaogullar:
Added A55 and A72 support (please excuse the force-pushes).
  
Note on `MINT` fix: I removed `MINT1 / MINT0` and made `MINT()` return `[SCALAR_MINT]`. Can't use just `MINT` because Python's `Enum` doesn't allow an enum member and a classmethod to share the same name. Followed the same pattern as the N1 model which uses `SCALAR_M` for the value and `M()` for the classmethod.
  
Let me know if you have any further comments.

slothy/targets/aarch64/cortex_a55.py
```diff
@@ -544,6 +562,7 @@ def get_min_max_objective(slothy):
     fmov_s_form: 1,  # from double/single to gen reg
     vdup_w: 1,
     (sxtb, uxtb): 1,
+    (crc32b, crc32h, crc32w, crc32x, crc32cb, crc32ch, crc32cw, crc32cx): 2,
```
mkannwischer: This should be 1. This is a common mistake when extending the SLOTHY models. 
Let me explain: The SWOG has the throughput over all execution units, while SLOTHY models the **inverse throughput per execution unit**. In other words, we model how long each execution unit is busy with such an instruction.
Since two execution units can execute this instruction and the overall throughput is 2, the inverse throughput per execution unit is 1.

Apologies if my comment was unclear.

mkannwischer (changes requested):
Awesome! Thanks for the quick turnaround and the changes.
Except for one mistake in the A55 model, this looks great to me.

slothy/targets/aarch64/cortex_a55.py
ckocaogullar: Fixed. Thanks for the explanation, that makes sense :)

mkannwischer (approved):
Thanks @ckocaogullar for the contribution and for making the changes.
Looks great to me now. 

Will merge once CI is green.

PR #442, opened 2026-05-08
AArch64: add fmov <Xd>, <Dn> support to A55, A72, and N1 models

ckocaogullar:
Add the 64-bit FP/SIMD scalar to GPR transfer (`fmov Xd, Dn`) to the architecture model and all three microarchitecture models.
  
  ### Scheduling data
  
  | Target | Unit | Lat | Tput | SWOG row | LLVM confirmation |
  |--------|------|-----|------|----------|-------------------|
  | N1 | V1 | 2 | 1 | "FP transfer, from vec to gen reg" | `WriteFCopy = N1Write_2c_1V1` |
  | A72 | L | 5 | 1 | "FP transfer, from vec to gen reg" | `WriteFCopy = A57Write_5cyc_1L` |
  | A55 | VEC | 1 | 1 | "FP transfer, from double/single to gen reg" | `WriteFCopy = CortexA55UnitFPALU` |
  
  ### Notes
  - `Dn` is a "SIMD and floating-point scalar register" (Arm ARM DDI0487G, Table C1-3). The SWOGs don't explicitly list which assembly forms fall under each row, but LLVM maps `FMOVDXr` to `WriteFCopy` on all three targets, matching the "FP transfer" rows above.
  - A55 SWOG throughput is listed as 2 (across 2 pipes); SLOTHY models inverse throughput per execution unit, so this becomes 1.
  
  Also adds the instruction to the shared `instructions.s` test.

### DIFF
--- a/slothy/targets/aarch64/aarch64_neon.py
+++ b/slothy/targets/aarch64/aarch64_neon.py
@@ -4043,6 +4043,12 @@ class fmov_s_form(Fmov):
     outputs = ["Wd"]
 
 
+class fmov_d_form(Fmov):
+    pattern = "fmov <Xd>, <Da>"
+    inputs = ["Da"]
+    outputs = ["Xd"]
+
+
 class fmov_0(Fmov):
     pattern = "fmov <Dd>, <Xa>"
     inputs = ["Xa"]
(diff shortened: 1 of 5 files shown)


mkannwischer (approved):
Thanks @ckocaogullar - looks excellent. I double checked the A55, A72, N1 SWOGs that there is indeed no difference between the s-form and the d-form.

PR #451, opened 2026-06-10
fix: don't discard optimized code when post-split perf estimate fails

### LINKED ISSUE #96: Don't abort if split_estimtate_perf fails
Right now, if the performance estimation after the splitting heuristic is applied fails, a `slothy.core.core.SlothyException: No solution found` exception is raised and not handled. Therefore, all the optimization process is gone. 

I think there should be a warning in the output but still, the optimized code should be written.

thisisjube:
The performance estimate after the split heuristic runs another binary
search (`optimize_binsearch`), which raises `SlothyException` on failure
rather than returning `success=False`. That exception propagated up and
discarded the already-optimized code. It also inherited the optimization
timeout, so a too-short timeout would abort it.

Catch the exception (warn and keep the optimized code without a stall
estimate) and disable the timeout for the estimate, since it only inserts
stalls into the already-fixed ordering.

Closes #96 @dop-amin

### DIFF
--- a/slothy/core/heuristics.py
+++ b/slothy/core/heuristics.py
@@ -961,14 +961,21 @@ class Heuristics:
             conf2.constraints.allow_renaming = False
             conf2.constraints.allow_reordering = False
             conf2.variable_size = True
-            stall_res = Heuristics.optimize_binsearch(
-                res.code, logger.getChild("split_estimtate_perf"), conf2
-            )
-            if stall_res.success is False:
-                log.error(
-                    "Stall-estimate for final code after split heuristic failed"
-                    " -- should not happen? Maybe increase timeout?"
-                    " Just returning the result without stall-estimate."
+            conf2.timeout = None
+            conf2.retry_timeout = None
+            stall_res = None
+            try:
+                stall_res = Heuristics.optimize_binsearch(
+                    res.code, logger.getChild("split_estimate_perf"), conf2
+                )
+            except SlothyException:
+                pass
+
+            if stall_res is None or stall_res.success is False:
+                log.warning(
+                    "Stall-estimate for final code after split heuristic"
+                    " failed -- returning the optimized result without"
+                    " stall-estimate."
                 )
             else:
                 res2 = Result(conf2)


slothy/core/heuristics.py
```diff
@@ -961,15 +961,25 @@ def not_empty(x):
...
             conf2.variable_size = True
-            stall_res = Heuristics.optimize_binsearch(
-                res.code, logger.getChild("split_estimtate_perf"), conf2
-            )
-            if stall_res.success is False:
-                log.error(
-                    "Stall-estimate for final code after split heuristic failed"
-                    " -- should not happen? Maybe increase timeout?"
-                    " Just returning the result without stall-estimate."
+            conf2.timeout = None
```
dop-amin, line 964: There is also `retry_timeout` that is used if a secondary objective is set that would trigger here. Should also be disabled.

slothy/core/heuristics.py
```diff
@@ -961,15 +961,25 @@ def not_empty(x):
...
+            stall_res = None
+            try:
+                stall_res = Heuristics.optimize_binsearch(
+                    res.code, logger.getChild("split_estimtate_perf"), conf2
+                )
+            except SlothyException:
```
dop-amin, line 971: Having almost the same warning twice is a bit redundant. 

What do you think about something like:
```python
  stall_res = None
  try:
      stall_res = Heuristics.optimize_binsearch(
          res.code, logger.getChild("split_estimtate_perf"), conf2
      )
  except SlothyException:
      pass

  if stall_res is None or stall_res.success is False:
      log.warning(
          "Performance estimate for final code after split heuristic"
          " failed -- returning the optimized result without stall-estimate."
      )
  else:
      res2 = Result(conf2)
      # ...
```

dop-amin (changes requested):
Thanks for the PR! 
Functionally, the secondary objective timeout should also be un-set.
From a style point of view, I think getting rid of the duplicate warning and simplifying the code path can be worth it.

slothy/core/heuristics.py
thisisjube: Ah, didn't know about this!

dop-amin (approved):
Looks good, thanks!

PR #454, opened 2026-06-14
RFC: Estimated-performance regression tests

mkannwischer:
Our tests/naive suite only checks that SLOTHY runs, not that the result is any good - so a regression that silently degrades a schedule would pass CI. This RFC proposes pinning SLOTHY's own performance estimate (predicted cycles/stalls) for a given snippet: since that estimate is the optimum of a minimization, it's deterministic and safe to assert exactly, turning "SLOTHY now thinks it found worse code" into a test failure. To enable it, optimize()/optimize_loop() now expose the result on slothy.last_result.

Concretely, a new pytest suite under tests/pytest/ reads per-case YAML sidecars - an assembly snippet plus the expected estimate per target (arch derived from target) and how to invoke SLOTHY - so adding a case is just dropping two files. 

I drafted this for discussion and would like to hear your opinions. Is this going to be useful in finding regressions in SLOTHY/ortools in the future? If so we can easily scale this up to the full instruction set supported. 

@dop-amin @hanno-becker


tests/pytest/estimated_performance/test_estimated_performance.py
```diff
@@ -0,0 +1,180 @@
...
+    slothy = Slothy(arch, target, logger=logger)
+    slothy.load_source_from_file(str(source_path))
+
+    # variable_size lets SLOTHY minimize stalls; a case may override it.
+    cfg = {"variable_size": True}
+    cfg.update(case.get("config") or {})
+    _apply_config(slothy.config, cfg)
+
+    # On Apple M1, x18 is reserved by the platform ABI.
+    if "m1" in target_name:
```
dop-amin, line 108: It's outside the scope of this PR but do you agree that we should have a mechanism to override arch model defaults from the uArch model? If so, I'd open an issue for that. 
Setting this explicitly all the time feels brittle.

tests/pytest/estimated_performance/test_estimated_performance.py
```diff
@@ -0,0 +1,180 @@
...
+    source_path = case_path.parent / source
+
+    logger = logging.getLogger(f"estperf.{case_path.stem}.{target_name}")
+    logger.setLevel(logging.WARNING)
+
+    slothy = Slothy(arch, target, logger=logger)
+    slothy.load_source_from_file(str(source_path))
+
+    # variable_size lets SLOTHY minimize stalls; a case may override it.
+    cfg = {"variable_size": True}
```
dop-amin, line 103: Possibly have a default config as yaml as well to make it more transparent?

dop-amin:
I think the feature itself will be useful and the implementation looks fine. 

Kind of related to #277 and #226.
Do you think it would be worth to think about a more general way to save the config alongside an optimization run once more? 

Could also consider to hook this into tests/examples so we will not have three separate places in which we keep track of "examples". Also, inside examples/tests it would be nice to move the config away from the Python script into yaml files. This would make them much more easily readable and not as convoluted.

PR #462, opened 2026-07-22
arm_v81m: parse Thumb width suffixes

bremoran:


mkannwischer (approved):
LGTM. Thanks @bremoran. 
It's does not add the width suffix to all instructions that allow it, but we can achieve completeness in a follow-up PR.

